PR #4, opened 2015-02-26
For #25503 - Changes and additions for Workfiles v2

alandann:
This pull request tidies up some existing code (including renaming files for consistency, etc. - sorry!) and adds a bunch of shared widgets implemented as part of the Workfiles v2 work.  

The main changes are:

- shotgun_view.WidgetDelegate now derives from a base implementation in views.WidgetDelegate so that the delegate can be used without the edit-on-selection mechanism.
- WidgetDelegate interface has been extended with separate get_paint_widget() & create_editor_widget() methods to allow more fine grained control in derived classes.  The original create_widget() is still supported for backwards compatibility
- Added a new ShotgunModelOverlayWidget that handles the connections to the shotgun model internally so that it's possible easily connect multiple overlay widgets to the same model.
- Added a hierarchical filtering proxy model that handles searching/filtering across a hierarchical model in a generic way

The new widgets added are:

__SpinnerWidget__
image
A simple animated widget used for displaying that something is busy.  This could be extended in the future to support an icon making it simple to replace the animation code in the overlay widget

__SearchWidget__
image
A simple widget containing a text field, icon and clear button with signals that get emitted when the search text changes.

__GroupedListView__
image
This is a custom view that presents a hierarchical model as a grouped list of items - the root level are the groups, the second level are the items within groups and all lower levels are ignored.


python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -0,0 +1,158 @@
...
+
+        :param src_row:         The row in the source model to filter
+        :param src_parent_idx:  The parent index in the source model to filter
+        :returns:               True if the row should be accepted by the filter, False
+                                otherwise
+        """
+        reg_exp = self.filterRegExp()
+        if self._filter_dirty or reg_exp != self._cached_regexp:
+            # clear the cache as the search filter has changed
+            self._accepted_cache = {}
```
jfboismenu-adsk: Not a big deal, but a _clear_cache method shared between this method and __init__ that takes the regex and the filter flag would have been nice. 4 things to keep track of when clearing the cache sounds like a maintenance nightmare when the cache is cleared more often or items are added to the cache.

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,150 @@
...
+        h_layout.setContentsMargins(3, 0, 3, 0)
+        h_layout.setSpacing(0)
+        self._ui.search_edit.setLayout(h_layout)
+
+        # hook up the signals:
+        self._ui.search_edit.textEdited.connect(self._on_text_edited)
+        self._ui.search_edit.returnPressed.connect(self._on_return_pressed)
+        self._clear_btn.clicked.connect(self._on_clear_clicked)
+                
+    # @property
```
jfboismenu-adsk: Did you forget to remove the comment?

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,150 @@
...
+    def _get_search_text(self):
+        """
+        get the search text from the widget
+        """
+        text = self._ui.search_edit.text()
+        return self._safe_to_string(text)
+
+    # @search_text.setter
```
jfboismenu-adsk: Same comment as above.

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,150 @@
...
+        self._clear_btn.setVisible(bool(text))
+        self.search_edited.emit(text)
+        
+    def _on_return_pressed(self):
+        """
+        Slot triggered when return has been pressed
+        """
+        self.search_changed.emit(self.search_text)
+        
+    def _safe_to_string(self, value):
```
jfboismenu-adsk: This looks like a utility that should be present in core or one of the frameworks.

python/search_widget/ui/search_widget.py
```diff
@@ -0,0 +1,40 @@
...
+    def setupUi(self, SearchWidget):
+        SearchWidget.setObjectName("SearchWidget")
+        SearchWidget.resize(341, 44)
+        self.horizontalLayout = QtGui.QHBoxLayout(SearchWidget)
+        self.horizontalLayout.setSpacing(0)
+        self.horizontalLayout.setContentsMargins(0, 0, 0, 0)
+        self.horizontalLayout.setObjectName("horizontalLayout")
+        self.search_edit = QtGui.QLineEdit(SearchWidget)
+        self.search_edit.setMinimumSize(QtCore.QSize(0, 24))
+        self.search_edit.setStyleSheet("#search_edit {\n"
```
jfboismenu-adsk: Why not a multi-line comment?

python/shotgun_view/widget_delegate.py
```diff
@@ -0,0 +1,141 @@
...
+        :param style_options: QStyleOptionViewItem object containing specifics about the 
+                              view related state of the cell.
+        """
+        pass        
+        
+    ########################################################################################
+    # 'private' methods that are not meant to be subclassed or called by a deriving class.
+        
+    def _on_selection_changed(self, selected, deselected):
+        """
```
jfboismenu-adsk: missing documentation for the parameters.

python/shotgun_view/widget_delegate.py
```diff
@@ -0,0 +1,141 @@
...
+        called when an "editor" is set up - the editor is set up 
+        via the openPersistentEditor call and is created upon selection
+        of an item.
+        
+        Normally, for performance, when we draw hundreds of grid cells, 
+        we use the same Qwidget as a brush and simply use it to paint.
+        
+        For the currently selected cell however, we need to be able to interact
+        with the widget (e.g. click a button for example) and therefore we need
+        to have a real widget for this.
```
jfboismenu-adsk: Same comment about parameters.

python/spinner_widget/spinner_widget.py
```diff
@@ -0,0 +1,133 @@
...
+        # call the base paint event:
+        QtGui.QWidget.paintEvent(self, event)
+        
+        painter = QtGui.QPainter()
+        painter.begin(self)
+        try:
+            painter.setRenderHint(QtGui.QPainter.Antialiasing)
+            
+            # TODO - replace with widget foreground color
+            pen = QtGui.QPen(QtGui.QColor(200, 200, 200))
```
jfboismenu-adsk: Is that something you forgot to do or couldn't do?

python/overlay_widget/shotgun_model_overlay_widget.py
```diff
@@ -0,0 +1,107 @@
...
+easily created and connected to the same shotgun model.
+"""
+
+from sgtk.platform.qt import QtCore, QtGui
+
+# load resources
+from .ui import resources_rc
+from .shotgun_overlay_widget import ShotgunOverlayWidget
+
+class ShotgunModelOverlayWidget(ShotgunOverlayWidget):
```
manneohrstrom: GOF say *prefer Composition over inheritance*. Just double checking here since this doesn't seem clear cut. Should this maybe be a HAS-A rather than a IS-A?

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,150 @@
...
+
+from .ui.search_widget import Ui_SearchWidget
+
+class SearchWidget(QtGui.QWidget):
+    """
+    Search widget class
+    """
+    
+    # emited when the search QTextField is being edited
+    search_edited = QtCore.Signal(object)# search text
```
manneohrstrom: Why `object` and not `str`? I am sure there is a good reason but a small comment explaining would be cool!

manneohrstrom:
Hey Alan! This is pretty exciting stuff! So nice with some more to re-use! :sunglasses: 

One quick question I had was about the `models` and `views` folders vs the existing `shotgun_view` folder.  Maybe it should be merged in some way? There are also two `widget_delegate` files - maybe these should be moved into their own folder? Curious to hear your thoughts!

alandann:
Hmm, yeah - my original thinking was that the shotgun_view stuff relied on the shotgun_model code but looking at it now that's not the case.  How about we move everything into a view module, drop the shotgun_ and then just provide the mapping in the __init__.py to shotgun_view for backwards compatibility?  For the widget delegate files, renaming them both to more appropriate names in a backwards compatible way seems the way to go - then they can both live in the view module?

alandann:
ok, this one is back for another round of code review - there have been a lot of changes and I've also tidied up the structure removing the loader specific widgets that used to live here.

There have been a significant number of changes so probably worth starting code-review from scratch, sorry!

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        Cached 'accepted' values for indexes.  Uses a dictionary that maps a key to a tuple 
+        containing a QPersistentModelIndex for the index and its accepted value.
+
+            key -> (QPersistentModelIndex, accepted)
+
+        In recent versions of PySide, the key is just a QPersistentModelIndex which has the
+        advantage that cache entries don't become invalid when rows are added/moved.  
+
+        In older versions of PySide this isn't possible as QPersistentModelIndex isn't hashable
```
jfboismenu-adsk: Which versions?

jfboismenu-adsk: What engines are affected?

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -0,0 +1,468 @@
...
+
+    def enable_caching(self, enable=True):
+        """
+        Allow control over enabling/disabling of the accepted cache used to accelerate
+        filtering.  Can be used for debug purposes to ensure the caching isn't the cause
+        of incorrect filtering/sorting or instability!
+
+        :param enable:    True if caching should be enabled, False if it should be disabled. 
+        """
+        self._dirty_all_accepted()
```
jfboismenu-adsk: A comment explaining why the entries become dirty just because the you are disabling or enabling caching would be appreciated.

python/navigation/navigation_widget.py
```diff
@@ -0,0 +1,120 @@
...
+class NavigationWidget(QtGui.QWidget):
+    """
+    Implementation of the NavigationWidget class
+    """
+
+    # Signal emitted whenever the user navigates using 
+    # the 'previous'/[<] or 'next'/[>] buttons.
+    navigate = QtCore.Signal(object)# destination to navigate to
+
+    # Signal emitted whenever the 'home'/[H] buttons is
```
jfboismenu-adsk: button

python/navigation/navigation_widget.py
```diff
@@ -0,0 +1,120 @@
...
+    # pressed
+    home_clicked = QtCore.Signal()
+
+    class _DestinationInfo(object):
```
jfboismenu-adsk: We should see if we still require support for Python 2.5. This could be replaced by a simple [namedtuple]( https://docs.python.org/2/library/collections.html#collections.namedtuple).

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,158 @@
...
+        self._ui.setupUi(self)
+        self.set_placeholder_text("Search")
+
+        # dynamically create the clear button so that we can place it over the
+        # edit widget:
+        self._clear_btn = QtGui.QPushButton(self._ui.search_edit)
+        self._clear_btn.setFocusPolicy(QtCore.Qt.StrongFocus)
+        self._clear_btn.setFlat(True)
+        self._clear_btn.setCursor(QtCore.Qt.ArrowCursor)
+        style = ("QPushButton {"
```
jfboismenu-adsk: Why isn't the style set on the button in the .ui file?

python/search_widget/search_widget.py
```diff
@@ -0,0 +1,158 @@
...
+        text = self._ui.search_edit.text()
+        return self._safe_to_string(text)
+    # @search_text.setter
+    def _set_search_text(self, value):
+        """
+        set the search text on the widget
+        """
+        self._ui.search_edit.setText(value)
+        self._clear_btn.setVisible(bool(value))
+    search_text = property(_get_search_text, _set_search_text)
```
jfboismenu-adsk: Any advantage to using this way or defining properties (apart from being able to access the private method directly in your class)?

python/spinner_widget/spinner_widget.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # public properties:
+        self.fps = 20
+        self.border = 2
+        self.line_width = 2
+        self.arc_length = 300
+        self.seconds_per_spin = 3
+
+        # timer used to force a repaint:
+        self._timer = QtCore.QTimer(self)
```
jfboismenu-adsk: Did you consider having a single timer event for all your spinners? The timer would go active when the first slot is connected and go inactive when the last slot is disconnected. I'm wondering if this could turn into a notification storm if there are a lot of spinners.

jfboismenu-adsk: To be clear, I agree with the current implementation, let's not overcomplicate things at first. I was simply curious.

python/views/grouped_list_view/grouped_list_view.py
```diff
@@ -0,0 +1,1070 @@
...
+            self._prev_viewport_sz = viewport_sz 
+
+        if not self._update_some_item_info and not self._update_all_item_info and not viewport_resized:
+            # nothing to do!
+            return
+
+        #print "%s, %s, %s, %s" % (self._update_all_item_info, self._update_some_item_info, viewport_resized, self._item_info)
+        self._update_all_item_info = self._update_all_item_info or viewport_resized
+
+        # (AD) TEMP
```
jfboismenu-adsk: What is temp?

alandann:
Addressed all code review and merged with master.  Any further changes can be made there!

PR #5, opened 2015-09-24
[minor] No default None value for ShotgunOverlayWidget parent (ctor)

jesseemond-adsk:
Since we call `parent.installEventFilter` right after, this will fail if the constructor had a default value. Clarifies that a value is really expected and needed.


manneohrstrom:
I did the exact same change in #6 as i was going through the docs. I'll close this one for now! Cheers!

PR #6, opened 2015-10-06
Added Shotgun Panel widgets and Sphinx docs.

manneohrstrom:
Tickets: #30634 #30645 #30636 #30637 #30638

- Adds sphinx documentation to Qt Widgets Fw
- Adds documentation to the workfiles2 additions
- Adds shotgun panel widgets + docs


python/version_label/version_label.py
```diff
@@ -0,0 +1,121 @@
...
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from sgtk.platform.qt import QtCore, QtGui
+
+from .ui import resources_rc
+
+class VersionLabel(QtGui.QLabel):
```
josh-t: This class is called **VersionLabel**, but it appears to accept any type of shotgun data. Perhaps it should only accept **Version** type data and bail if given something else?

python/version_label/version_label.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        self._sg_data = sg_data
+        
+        # based on the data, figure out if the icon should be active or not
+        self._active = False
+        
+        if sg_data and sg_data.get("type") == "Version":
+            # versions are supported
+            if sg_data.get("sg_uploaded_movie"):
+                self._active = True
```
josh-t: Since the logic for determining whether the data is playable is in this class, perhaps there should be a `playable` property exposed for use by the calling code. Having that property available would allow the user to selectively connect to the `playback_clicked` signal rather than assuming the **Version** data is playable. 

I'd probably recommend changing `_active` to `_playable` as well.

python/version_label/version_label.py
manneohrstrom: discussed with @josh-t, changing to `ShotgunPlaybackLabel`

python/version_label/version_label.py
manneohrstrom: fixed!

python/note_input_widget/note_editor.py
```diff
@@ -0,0 +1,363 @@
...
+        # go forwards to find end pos. This is relevant if the cursor is 
+        # not positioned at the end of the line
+        while end_idx < len(full_text) and full_text[end_idx] not in (" ", "\n"):
+            end_idx += 1
+            
+        current_word = full_text[start_idx+1:end_idx]
+        return current_word
+
+
+    def _do_sg_global_search(self, sg, data):
```
josh-t: Curious why this is a dictionary and not a string.

josh-t:
One small question for NoteEditor, otherwise looks awesome!

python/screen_grab/screen_grab.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        Key press event
+        """
+        # for some reason I am not totally sure about, it looks like
+        # pressing escape while this dialog is active crashes Maya.
+        # I tried subclassing closeEvent, but it looks like the crashing
+        # is triggered before the code reaches this point.
+        # by sealing the keypress event and not allowing any further processing
+        # of the escape key (or any other key for that matter), the 
+        # behaviour can be successfully avoided.
```
josh-t: Would it make sense to process Esc key press here and manually close the window?

python/screen_grab/screen_grab.py
```diff
@@ -0,0 +1,211 @@
...
+        desktop.screenCountChanged.connect(self._fit_screen_geometry)
+
+    @property
+    def capture_rect(self):
+        """
+        The resulting QRect from a previous capture operation.
+        """
+        return self._capture_rect
+
+    def paintEvent(self, evt):
```
josh-t: just a personal thing, but i'd prefer `event` over `evt` throughout this class.

python/global_search_widget/global_search_widget.py
```diff
@@ -0,0 +1,284 @@
...
+        self.__sg_data_retriever.clear()
+        
+        # clear thumbnail map
+        self._thumb_map = {}
+        
+        # kick off async data request from shotgun 
+        data = {"text": text}
+        self._processing_id = self.__sg_data_retriever.execute_method(self._do_sg_global_search, data)        
+        
+    def _do_sg_global_search(self, sg, data):
```
josh-t: Same question as note editor. Just wondering if `data` needs to be a dict. Is there other data that may go in here?

python/note_input_widget/note_editor.py
manneohrstrom: Ahh, this is because the `self.__sg_data_retriever.execute_method(self._do_sg_global_search, data)` call. The `execute_method()` call takes a function pointer and a dictionary as inputs. These are then passed at execution time (in a worker thread) to the method - in this case `_do_sg_global_search()`. I just used a dictonary to keep things simple - so that all options to the worker method are passed in a single data structure. That seemed easy to serialize, print out in logging etc. Hope it makes sense!

I'll add some comments to clarify!

python/global_search_widget/global_search_widget.py
manneohrstrom: added comments to clarify!

python/screen_grab/screen_grab.py
manneohrstrom: I tried doing this, but wasn't successful - for some reason it crashes in Maya. I didn't have time to investigate so I just "plugged the whole" with the fix above. Admittedly, I didn't dig very deep - just trying some obvious stuff - and it would be nice to be able to hit ESC to cancel - that's how the macosx screen grabber work for example. At the same time, it feels like a nice to have! :) Unless you object, I suggest put a ticket in so we can potentially address it later. What do you think?

python/screen_grab/screen_grab.py
manneohrstrom: fixed!

python/activity_stream/activity_stream.py
```diff
@@ -0,0 +1,552 @@
...
+        try:
+
+            # we are building the widgets bottom up.
+            # first of all, insert a widget that will expand so that 
+            # it consumes all unused space. This is to keep other
+            # widgets from growing when there are only a few widgets
+            # available in the scroll area.
+            self._bundle.log_debug("Adding expanding base widget...")            
+            
+            expanding_widget = QtGui.QLabel(self)
```
josh-t: Any advantage to this over calling `addStretch()` on the layout?

python/activity_stream/activity_stream.py
```diff
@@ -0,0 +1,552 @@
...
+                self.ui.activity_stream_layout.removeWidget(w)
+                w.setParent(None)
+                w.deleteLater()
+            self._other_widgets = []
+        
+        finally:
+            # make the window visible again and trigger a redraw
+            self.setVisible(True)
+            
+    def _clear_loading_widget(self):
```
josh-t: Would it make sense to have a persistent loading widget that doesn't live in the activity stream, but just above it in the layout. As opposed to deleting and recreating?

python/activity_stream/activity_stream.py
```diff
@@ -0,0 +1,552 @@
...
+        # (which only happens on a full load)
+        self.__overlay.hide()
+            
+    def _process_thumbnail(self, data):
+        """
+        New thumbnail has arrived from the data manager
+        """
+        # broadcast to all activity widgets
+        for widget in self._widgets.values():
+            widget.set_thumbnail(data)
```
josh-t: At first glance, this was confusing to me. Why would we set the same thumbnail for all widgets? But I think this is forwarding thumbnail data to all the widgets in case they need to display it? Not a huge thing, but maybe a verb other than `set`? Maybe `apply`?

python/activity_stream/activity_stream.py
```diff
@@ -0,0 +1,552 @@
...
+        # set up signals
+        self._data_manager.note_arrived.connect(self._process_new_note)
+        self._data_manager.update_arrived.connect(self._process_new_data)
+        self._data_manager.thumbnail_arrived.connect(self._process_thumbnail)
+        self.ui.note_widget.data_updated.connect(self._on_note_submitted)
+        
+        # keep handles to all widgets to be nice to the GC
+        self._loading_widget = None
+        self._other_widgets = []
+        self._widgets = {}
```
josh-t: I think these are just activity stream widgets? Could we call this `_stream_widgets` or `_as_widgets` maybe? Then `_other_widgets` would be a little clearer too.

python/activity_stream/data_manager.py
```diff
@@ -0,0 +1,838 @@
...
+            c = None
+            raise
+
+        finally:
+            if c:
+                c.close()
+
+        return connection
+ 
+    def __get_note_thread_data(self, note_id):
```
josh-t: These db methods all have the same try/except/finally structure. Perhaps a context manager to get the cursor and do the cleanup for you?

python/activity_stream/label_widgets.py
```diff
@@ -0,0 +1,206 @@
...
+        
+    def mousePressEvent(self, event):
+        """
+        Fires when the mouse is pressed
+        """
+        QtGui.QLabel.mousePressEvent(self, event)        
+        self.clicked.emit()
+    
+    
+class LargeAttachmentThumbnail(ClickableTextLabel):
```
josh-t: Maybe the base class should just be ClickableLabel?

python/activity_stream/label_widgets.py
```diff
@@ -0,0 +1,206 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+"""
+A collection of various subclassed QLabels used by the activity stream UI
+"""
+from sgtk.platform.qt import QtCore, QtGui
+from . import utils
+import sgtk
+
+class ClickableTextLabel(QtGui.QLabel):
```
josh-t: Curious why not use a button class with a flat styling?

python/activity_stream/reply_list.py
```diff
@@ -0,0 +1,415 @@
...
+            # set it's parent to None so that it is removed from the widget hierarchy
+            x.setParent(None)
+            # mark it to be deleted when event processing returns to the main loop
+            x.deleteLater()
+    
+        self._general_widgets = []
+        self._reply_widgets = []
+        self._attachment_group_widgets = {}    
+        
+    def _load_stylesheet(self):
```
josh-t: I've seen this method a few places. Looks like it could be a utility method somewhere.

python/activity_stream/widget_activity_stream_base.py
```diff
@@ -0,0 +1,220 @@
...
+            
+    def _icon_url_for_entity_type(self, entity_type):
+        """
+        Given an entity type, return a dark 16x16px icon resource url.
+        Returns None if no icon exists
+        
+        :param entity_type: entity type
+        :returns: resource url or None
+        """
+        entity_types_with_icons = ["Asset",
```
josh-t: This list might be useful in other UIs we build.

python/activity_stream/widget_activity_stream_base.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        try:
+            (entity_type, entity_id) = url.split(":")
+            entity_id = int(entity_id)
+            self.entity_requested.emit(entity_type, entity_id)
+        except:
+            self._bundle.log_warning("Could not parse url '%s'" % url)
+
+        
+    def _set_timestamp(self, data, label):
```
josh-t: This logic is also really handy for other UIs.

python/activity_stream/widget_attachment_group.py
```diff
@@ -0,0 +1,139 @@
...
+            obj.clicked.connect(self._toggle_large_thumbnails)            
+            self.ui.preview_layout.addWidget(obj, current_row, current_col)
+            self._small_thumbnails[data["id"]] = obj
+            
+            
+            if current_col > 4:
+                current_col = 0
+                current_row += 1
+            else:
+                current_col += 1
```
josh-t: What was the reason not to use something like `QGridLayout`?

python/activity_stream/widget_note.py
```diff
@@ -0,0 +1,234 @@
...
+                self._reply_widgets.append(w)
+                # ensure navigation requests from replies bubble up
+                w.entity_requested.connect(self.entity_requested.emit)
+                # next bunch of attachments will be after a reply
+                # rather than directly under the note
+                # (this affects the visual style) 
+                attachment_is_directly_after_note = False
+
+                
+
```
josh-t: extra blank lines here.

python/elided_label/elided_label.py
```diff
@@ -34,11 +34,12 @@ def __init__(self, parent=None):
     #@property
```
josh-t: This looks like it should be removed.

josh-t:
Activity stream CR. Overall, looks awesome! A few minor questions/suggestions.

python/screen_grab/screen_grab.py
manneohrstrom: Added ticket 31755 to track this.

python/activity_stream/activity_stream.py
manneohrstrom: I picked a label to make sure that it wouldn't affect the external boundaries. Mostly because I knew this would work and wouldn't require lots of tweaking. A better be safe than sorry approach I guess. Do you think it's terrible? Maybe I should have used `QSpacerItem` with policy preferred?

josh-t:
External screen grab CR looks good. Couple of minor comments. Back to IP.

python/activity_stream/activity_stream.py
manneohrstrom: I think the reason for this is that it needs to be part of the scroll area, so that you can scroll down while it is loading. I'll add a comment to clarify.

python/activity_stream/activity_stream.py
manneohrstrom: Yes, this is confusing and needs cleaning up. :(

python/activity_stream/activity_stream.py
manneohrstrom: renamed to `self._activity_stream_static_widgets` and `self._activity_stream_data_widgets`

python/activity_stream/label_widgets.py
manneohrstrom: yes! fixed!

python/activity_stream/data_manager.py
manneohrstrom: great suggestion. replaced with decorator

python/activity_stream/label_widgets.py
manneohrstrom: if i remember correctly, i was having issues getting the text correctly centered and went for a solution that i knew would work. But i am sure it's possible to get it to work - and more elegant too...

python/activity_stream/widget_note.py
manneohrstrom: fixed!

python/activity_stream/widget_activity_stream_base.py
manneohrstrom: added this to shotgun_globals

python/activity_stream/widget_activity_stream_base.py
manneohrstrom: this is currently not shared. My thinking was that we can potentially share it at the point of reuse.

python/activity_stream/widget_attachment_group.py
manneohrstrom: it is actually using a gridlayout. Is there an easier/more elegant way to do this?

josh-t: hmmm. clearly i was confused when i made that comment. not sure what i thought was going on. looks good!

PR #8, opened 2015-12-08
For #32981, rebuild slowness in workfiles

jfboismenu:
Listens for an signal that the source model is about to be reset. If it is, it clears the cache as the indices are about to be all invalidated anyway.


manneohrstrom:
cr looks good!

PR #10, opened 2016-02-11
Generic field widgets to display arbitrary Shotgun data

robblau:
A set of generic widgets that can be used to display Shotgun fields and their data and a manager class to allow them to be generated from the schema info.  Read only right now.

An example of how they can be used is:

```python
    """
    Main application dialog window
    """

    def __init__(self):
        """
        Constructor
        """
        QtGui.QWidget.__init__(self)

        self._app = sgtk.platform.current_bundle()

        self._fields_manager = shotgun_fields.ShotgunFieldManager()
        self._fields_manager.initialized.connect(self.populate_dialog)
        self._fields_manager.initialize()

    def populate_dialog(self):
        fields = sorted(self._app.shotgun.schema_field_read("CustomEntity02").keys())
        entities = self._app.shotgun.find("CustomEntity02", [], fields=fields)

        field_labels = []
        for field in fields:
            field_labels.append(shotgun_globals.get_field_display_name("CustomEntity02", field))

        # set headers
        table = QtGui.QTableWidget(len(entities), len(fields), self)
        table.setHorizontalHeaderLabels(field_labels)

        # build a table for the entities
        for (i, entity) in enumerate(entities):
            for (j, field) in enumerate(fields):
                widget = self._fields_manager.create_widget("CustomEntity02", field, entity)
                if widget is None:
                    widget = QtGui.QLabel("No widget")
                table.setCellWidget(i, j, widget)

        table.resizeColumnsToContents()

        layout.addWidget(table)
        self.setLayout(layout)
```


manneohrstrom:
Looks like this would be easy to plug into the "all fields" view in the shotgun panel:
https://github.com/shotgunsoftware/tk-multi-shotgunpanel/blob/master/python/app/widget_all_fields.py

Cool!

manneohrstrom:
The big question for me is performance: with a table showing say 10 fields for 50 entities, how does it perform? We moved away a bit from full-on widget based approaches in the past because QT can get a little slow when the number of widgets grow. But maybe this is more of an issue around *how* you would use this system than about the system itself? I am also super curious to try implementing this in a MVC deletegate model - so that the "editor" mode in the delegate calls out to the field manager and requests the editor. That could be pretty sweet and potentially a fast way to enable field value editing in any app that is using the qt widget delegate stuff! :)

robblau:
The plan for using these in actual production is to have a standard delegate that understands how the widgets work so that it can work directly with the Shotgun model so that you'll only ever need to instantiate one widget of each field type that you are showing.  That delegate would also understand the editor widgets as well to manage switching from one to the other.

So yeah... this is just step one and the code above is just a good way to demo the functionality.  I'd definitely not use a table like that for any kind of serious data viewing.  The rest comes in a little bit.  :)

manneohrstrom:
I think this sounds super awesome and will help apps across the board! :+1: :star2:

python/shotgun_fields/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from . import shotgun_field_manager
+ShotgunFieldManager = shotgun_field_manager.ShotgunFieldManager
```
thebeeland: Any reason why this isn't:

from .shotgun_field_manager import ShotgunFieldManager

thebeeland: Any reason why this can't be:

from .shotgun_field_manager import ShotgunFieldManager

python/shotgun_fields/checkbox_widget.py
```diff
@@ -0,0 +1,57 @@
...
+from .shotgun_field_manager import ShotgunFieldManager
+
+
+class CheckBoxWidget(QtGui.QCheckBox):
+    """
+    Inherited from a :class:`~PySide.QtGui.QCheckBox`, this class is able to
+    display a checkbox field value as returned by the Shotgun API.
+    """
+
+    def __init__(self, parent=None, entity=None, field_name=None, bg_task_manager=None, **kwargs):
```
thebeeland: Might should be *args, **kwargs in case positional arguments are used across the board and include additional QCheckBox arguments.

python/shotgun_fields/checkbox_widget.py
```diff
@@ -0,0 +1,57 @@
...
+        :param entity: The Shotgun entity dictionary to pull the field value from.
+        :type entity: Whatever is returned by the Shotgun API for this field
+
+        :param field_name: Shotgun field name
+        :type field_name: String
+
+        :param bg_task_manager: The task manager the widget will use if it needs to run a task
+        :type bg_task_manager: :class:`~task_manager.BackgroundTaskManager`
+        """
+        QtGui.QCheckBox.__init__(self, parent, **kwargs)
```
thebeeland: If using line 24 comment above, then `*args, **kwargs` here.

python/shotgun_fields/currency_widget.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
thebeeland: Minor: 2016

python/shotgun_fields/date_widget.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Set the value displayed by the widget.
+
+        :param value: The value displayed by the widget
+        :type value: String in the form YYYY-MM-DD representing the date
+        """
+        if value is None:
+            self.clear()
+        else:
+            dt = datetime.datetime.strptime(value, "%Y-%m-%d")
```
thebeeland: Maybe make the tokenized format an optional argument with this as the default?

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        QtGui.QLabel.__init__(self, parent, **kwargs)
+        self.set_value(entity[field_name])
+
+    def set_value(self, value):
+        """
+        Set the value displayed by the widget.
+
+        :param value: The value displayed by the widget
+        :type value: :class:`~datetime.datetime`
```
thebeeland: That type seems incorrect. Isn't it expecting the field dict from the API?

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        if value is None:
+            self.clear()
+        else:
+            self.setText(self._create_human_readable_timestamp(value, " %I:%M%p"))
+
+    def _create_human_readable_timestamp(self, dt, postfix=""):
```
thebeeland: Might be nice to store the datetime object and provide a getter for it (or store it as a property). Then we can get the string representation in "pretty" form as well as something more useful from a code standpoint.

python/shotgun_fields/entity_widget.py
```diff
@@ -0,0 +1,78 @@
...
+            self.clear()
+        else:
+            self.setText(self._entity_dict_to_html(value))
+
+    def _entity_dict_to_html(self, value):
+        """
+        Translate the entity dictionary to html that can be displayed in a
+        :class:`~PySide.QtGui.QLabel`.
+
+        :param value: The entity dictionary to conver to html
```
thebeeland: "conver"

python/shotgun_fields/float_widget.py
```diff
@@ -0,0 +1,58 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Widget that represents the value of a float field in Shotgun
+"""
+import locale
```
thebeeland: I don't think locale is used.

python/shotgun_fields/image_widget.py
```diff
@@ -0,0 +1,155 @@
...
+        :param field_name: The name of the field the value is for
+        :type field_name: String
+        """
+        if value is None:
+            self.clear()
+            return
+
+        # queue up the download in the background
+        self._task_uid = self._data_retriever.request_thumbnail(
+                value,  # url must still be valid since we don't know the entity this is for
```
thebeeland: Minor: is double tab beyond "self" in line 77 here intentional?

python/shotgun_fields/list_widget.py
```diff
@@ -0,0 +1,26 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+Widget that represents the value of a list field in Shotgun
+"""
+from .shotgun_field_manager import ShotgunFieldManager
+from .label_base_widget import LabelBaseWidget
+
+
+class ListWidget(LabelBaseWidget):
```
thebeeland: Probably not worth worrying about, but should we be concerned with confusion related to ListWidget vs. QListWidget?

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        self.initialized.emit()
+
+    def supported_fields(self, sg_entity_type, field_names):
+        """
+        Returns the subset of fields from field_names that have an associated widget class.
+
+        :param sg_entity_type: Shotgun entity type
+        :type sg_entity_type: String
+
+        :param field_names: An array of Shotgun field names
```
thebeeland: Minor: s/array/list/

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+            # instantiate the widget
+            return cls(
+                parent=parent,
+                entity=entity,
+                field_name=field_name,
+                bg_task_manager=self._task_manager,
+                **kwargs
+            )
+        return None
```
thebeeland: Personal preference, but I would probably raise instead of returning None.

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        display_name = shotgun_globals.get_field_display_name(sg_entity_type, field_name)
+        return QtGui.QLabel(display_name)
+
+# import the actual field types to give them a chance to register
+from . import checkbox_widget, currency_widget, date_and_time_widget, date_widget
+from . import entity_widget, file_link_widget, float_widget, footage_widget
+from . import image_widget, list_widget, multi_entity_widget, number_widget
+from . import percent_widget, status_list_widget, tags_widget, text_widget
+from . import url_template_widget
```
thebeeland: Might be a little easier/cleaner to do this:

```
from . import (
    checkbox_widget,
    currency_widget,
    date_and_time_widget,
    ...
)
```

manneohrstrom:
Sorry if I'm late to the party! I wanted to wait for the refactor to happen around the metaclass stuff. I'll do a quick CR now!

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -0,0 +1,65 @@
...
+        delta = datetime.datetime.now(dt.tzinfo) - dt
+
+        if delta.days == 1:
+            format = "Yesterday%s" % postfix
+        elif delta.days == 0:
+            format = "Today%s" % postfix
+        elif delta.days == -1:
+            format = "Tomorrow%s" % postfix
+        else:
+            format = "%%x%s" % postfix
```
manneohrstrom: Please add comment here along the lines of
```
# use the date formatting associated with the current locale
```

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -0,0 +1,65 @@
...
+    _FIELD_TYPE = "date_time"
+
+    def _string_value(self, value):
+        """
+        Convert the Shotgun value for this field into a string
+
+        :param value: The value to convert into a string
+        :type value: :class:`datetime.datetime`
+        """
+        return self._create_human_readable_timestamp(value, " %I:%M%p")
```
manneohrstrom: Looking here: https://docs.python.org/2/library/time.html#time.strftime
Could `%X` be used instead so that the default locale is taken int account?

python/shotgun_fields/image_widget.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Set the value displayed by the widget.
+
+        :param value: The value displayed by the widget
+        :type value: A String that is a valid URL for the thumbnail image
+        """
+        # queue up the download in the background
+        entity_id = self._entity and self._entity.get("id") or None
+        entity_type = self._entity and self._entity.get("type") or None
+        self._task_uid = self._data_retriever.request_thumbnail(
```
manneohrstrom: if you pass a `request_thumbnail=True` parameter here, the data dictionary you get back when the worker signals will contain a ready-loaded `QImage` instance of the thumbnail. (docs here: http://developer.shotgunsoftware.com/tk-framework-shotgunutils/shotgun_data.html#shotgun_data.ShotgunDataRetriever.request_thumbnail ). This might be nice because the image is then loaded from disk and processed in the worker thread instead of in the main one!

manneohrstrom: example code here:
https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/master/python/global_search_widget/global_search_widget.py#L275

and here:
https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/master/python/global_search_widget/global_search_widget.py#L246

python/shotgun_fields/status_list_widget.py
```diff
@@ -0,0 +1,43 @@
...
+
+        :param value: The value to convert into a string
+        :type value: String that is a valid Shotgun status code
+        """
+        str_val = shotgun_globals.get_status_display_name(value)
+        color_str = shotgun_globals.get_status_color(value)
+
+        if color_str:
+            # append colored box to indicate status color
+            str_val = ("<span style='color: rgb(%s)'>&#9608;</span>&nbsp;%s" % (color_str, str_val))
```
manneohrstrom: we really need to get some vector images for those status icons, at least the "basic" set. Would be so nice!

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -0,0 +1,244 @@
...
+            # instantiate the widget
+            return cls(
+                parent=parent,
+                entity=entity,
+                field_name=field_name,
+                bg_task_manager=self._task_manager,
+                **kwargs
+            )
+
+        return None
```
manneohrstrom: maybe it would be more helpful to just fall back on a default QLabel (or something) in case the value cannot be realised in the framework?

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -0,0 +1,244 @@
...
+
+        :returns: :class:`~PySide.QtGui.QWidget` or None if the field type has no display widget
+        """
+        data_type = shotgun_globals.get_data_type(sg_entity_type, field_name)
+
+        if data_type in self.__FIELD_TYPE_MAP:
+            cls = self.__FIELD_TYPE_MAP[data_type]
+
+            # instantiate the widget
+            return cls(
```
manneohrstrom: The creation of the widget object and the setting of the value (via the `entity`) is now all handled in one go. Wondering if it might be cleaner to have to separate calls, one to construct the widget and then set the value? So basically first construct the widget, then call `obj.set_value()`? 

Without trying to guess too much, I suspect future additions to the `ShotgunFieldManager` may have the construction of widgets and the setting of values more decoupled (e.g. for delegates or keeping a pool of widgets around that are being reused or something).

manneohrstrom:
CR done! Added some minor comments and questions! Super exciting stuff!

I guess for the sphinx side of things, we don't need to worry too much as long as all expose in the API level documentation is the `ShotgunFieldManager` class? (curious to see what sphinx makes of the meta classes though!)

PR #12, opened 2016-02-17
Ticket #34755 - Support for building a fields menu for a given entity type

robblau:
Adds a new module that contains the EntityFieldMenu, a QMenu subclass that knows how to display a menu consisting of all the fields for a given Shotgun entity, including support for bubbled fields.

An example of using this menu is:

```python
        def __init__(self):
            QtGui.QWidget.__init__(self)

            # grab a field manager to know what fields are displayable
            self._field_manager = shotgun_fields.ShotgunFieldManager()

            # setup a label to have the fields menu as its context menu
            self.label = QtGui.QLabel("Right click me!")
            self.label.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
            self.label.customContextMenuRequested.connect(self.open_menu)

            # and layout the dialog
            layout.addWidget(self.label)
            self.setLayout(layout)

        def field_filter(self, field):
            # display fields that are displayable by the shotgun field widgets
            return bool(self._field_manager.supported_fields("CustomEntity02", [field]))

        def checked_filter(self, field):
            # none of the fields are checked
            return False

        def disabled_filter(self, field):
            # none of the fields are disabled
            return False

        def open_menu(self, position):
            menu = shotgun_menus.EntityFieldMenu("CustomEntity02")

            # attach our filters
            menu.set_field_filter(self.field_filter)
            menu.set_checked_filter(self.checked_filter)
            menu.set_disabled_filter(self.disabled_filter)

            # show the menu and print the result
            action = menu.exec_(self.label.mapToGlobal(position))
            if action:
                # action's data has the field that was selected
                self.do_thing(action.data()["field"])
```

Which will look like this

screen shot 2016-02-17 at 8 34 17 am


python/shotgun_menus/entity_field_menu.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        Add a label with the given title to the menu
+
+        :param title: The title of the sectional label
+        :type title: String
+        """
+        # build the label widget
+        label = QtGui.QLabel(self)
+        label.setStyleSheet("margin: 0.25em; color: gray;")
+        font_style = "font-variant: small-caps; font-weight: bold; font-size: large"
```
manneohrstrom: Personal opinon: the way QT renders the artificial small caps makes me feel slightly sick. `text-transform: uppercase` instead?

python/shotgun_menus/entity_field_menu.py
```diff
@@ -0,0 +1,311 @@
...
+
+class EntityFieldMenu(QtGui.QMenu):
+    """
+    Inherited from a :class:`~PySide.QtGui.QMenu`, this menu will automatically display the fields
+    for a given Shotgun entity.
+
+    The QActions for the menu will all have their data set to a dictionary in the form:
+        {"field": selected_field}
+    """
+    _AUDIT_FIELDS = ["created_by", "created_at", "updated_by", "updated_at"]
```
manneohrstrom: could this be useful to include in the globals?

python/shotgun_menus/entity_field_menu.py
```diff
@@ -0,0 +1,311 @@
...
+        if audit_fields:
+            self.addSeparator()
+            self.add_label("Audit Fields")
+            for field_info in audit_fields:
+                self._add_qaction(field_info["field"], field_info["name"])
+
+        # and finally bubble fields
+        if bubble_fields:
+            self.addSeparator()
+            self.add_label("Linked Fields")
```
manneohrstrom: Would be pretty nice if we could get the entity type icons from the shotgun globals to decorate the menu items for the linked fields! :)

python/shotgun_menus/entity_field_menu.py
robblau: @manneohrstrom do you have a use case in mind?  Easy enough to move there if needed, but unless you have something in mind I'd rather wait than create a dependency between this and a core update.

python/shotgun_menus/entity_field_menu.py
robblau: Agreed.  Done and MUCH better.

python/shotgun_menus/entity_field_menu.py
robblau: :)  I think that could come off as busy.  Plus it would be different than what is in Shotgun, which I'd avoid at this point.

robblau:
Closing this pull request to create one with the rv branch as the target.

PR #14, opened 2016-02-20
Ticket #34756 - Render delegates for shotgun model

robblau:
Adds support for displaying Shotgun field widgets directly in a view that is being populated by a ShotgunModel.

This code

```python
    def __init__(self):
        QtGui.QWidget.__init__(self)

        # grab a field manager to get the delegate
        self.fields_manager = shotgun_fields.ShotgunFieldManager()
        self.fields_manager.initialized.connect(self.on_initialized)
        self.fields_manager.initialize()

    def on_initialized(self):
        # grab all displayable fields on CustomEntity02, with "code" first
        fields = shotgun_globals.get_entity_fields("CustomEntity02")
        fields = self.fields_manager.supported_fields("CustomEntity02", fields)
        fields = ["code"] + [f for f in fields if f != "code"]

        # setup the model
        self.model = shotgun_model.SimpleShotgunModel(self)
        self.model.load_data("CustomEntity02", filters=[], fields=fields, columns=fields)

        # create the table view
        self.table = views.ShotgunTableView(self.fields_manager, parent=self)
        self.table.setModel(self.model)

        # and layout the dialog
        layout.addWidget(self.table)
        self.setLayout(layout)
```

will result in a dialog like this:

image


python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -219,6 +250,29 @@ def create_display_widget(self, sg_entity_type, field_name, entity=None, parent=
 
         return None
 
+    def create_delegate(self, sg_entity_type, field_name, view):
```
manneohrstrom: trying to think ahead about editing - trying to avoid making up use cases but wondering if there is a need for a flag to specify that a delegate should not be editable? Or is this perhaps a property that belongs to the role of the model item?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -0,0 +1,86 @@
...
+        :param model_index: object representing the data of the object that is
+                            about to be drawn.
+        :type model_index:  :class:`~PySide.QtCore.QModelIndex`
+
+        :param style_options: Object containing specifics about the
+                              view related state of the cell.
+        :type style_options:  :class:`~PySide.QtGui.QStyleOptionViewItem`
+        """
+        value = model_index.data(shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
+        sanitized_vlue = shotgun_model.sanitize_qt(value)
```
manneohrstrom: nit: typo

python/views/shotgun_tableview.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Overrides the base class setModel.  This assumes that the model is a ShotgunModel
+        and will set the delegates for the each column to the appropriate delegate to display
+        its Shotgun value.
+        """
+        QtGui.QTableView.setModel(self, model)
+
+        # set the delegates
+        columns_and_fields = model.get_additional_column_fields()
+        for (column, field) in columns_and_fields:
```
manneohrstrom: i think maybe saying `column_idx` makes this easier to read - but also see comment in the shotgun model about potentially returning more data as part of `get_additional_column_fields()`

manneohrstrom:
This is super exciting. And slots in very nicely with the stuff we already have. Minor CR comments only. good to go!

python/views/shotgun_tableview.py
```diff
@@ -0,0 +1,45 @@
...
+        :param parent: Parent widget
+        :type parent: :class:`PySide.QtGui.QWidget`
+        """
+        QtGui.QTableView.__init__(self, parent)
+        self._fields_manager = fields_manager
+
+    def setModel(self, model):
+        """
+        Overrides the base class setModel.  This assumes that the model is a ShotgunModel
+        and will set the delegates for the each column to the appropriate delegate to display
```
thebeeland: typo: s/for the each/for each/

python/shotgun_fields/shotgun_field_manager.py
robblau: That should be in the model:
https://doc.qt.io/qt-4.8/model-view-programming.html#making-the-model-editable

Definitely going to verify that when we get to editors for the Shotgun fields.

PR #16, opened 2016-03-25
#34765 note attachments

thebeeland:
Allows for multiple, arbitrary file attachments when creating notes and replies. The screencapture button is back and now triggers an RV event that will tell the Mode to perform RV's native screengrab behavior and then call back to the details panel with the path to the resulting image. The RV Mode portion of this has not yet been implemented (Toolkit won't handle that part).

When a new note is started:

new_note

When the "attach" button (the paperclip icon) is clicked and attachments are added then via the "plus" button:

with_attachments

results


python/activity_stream/activity_stream.py
```diff
@@ -102,7 +109,51 @@ def destroy(self):
...
+    def _set_allow_screenshots(self, state):
+        """
+        Sets whether to allow screenshots.
+
+        :param state: True or False
+        """
+        self._allow_screenshots = bool(state)
+        self.ui.note_widget.allow_screenshots(self._allow_screenshots)
+
+    allow_screenshots = QtCore.Property(
```
robblau: Not sure what the sphinx syntax is for something like a property, but seems like this should have some associated documentation.

python/activity_stream/activity_stream.py
```diff
@@ -102,7 +109,51 @@ def destroy(self):
...
+
+    def _set_show_sg_stream_button(self, state):
+        """
+        Sets whether to show the button to navigate to Shotgun.
+
+        :param state: True or False
+        """
+        self._show_sg_stream_button = bool(state)
+
+    show_sg_stream_button = QtCore.Property(
```
robblau: Docs.

python/activity_stream/ui/activity_stream_widget.py
```diff
@@ -1,44 +1,70 @@
 # -*- coding: utf-8 -*-
 
-# Form implementation generated from reading ui file 'activity_stream_widget.ui'
+# Form implementation ORIGINALLY generated from reading ui file 'activity_stream_widget.ui'
 #
 #      by: pyside-uic 0.2.13 running on PySide 1.1.1
 #
-# WARNING! All changes made in this file will be lost!
+# *******  NOT USING QT DESIGNER FOR THIS TEMPORARILY
```
robblau: Not sure what the status of this is... just flagging it as part of code review to make sure there isn't some cleanup that should happen at this point...

python/activity_stream/ui/note_widget.py
```diff
@@ -5,65 +5,96 @@
 #      by: pyside-uic 0.2.13 running on PySide 1.1.1
```
robblau: Seems like if we're not generating these files from the .ui files they shouldn't be in the ui directory.

python/note_input_widget/widget.py
```diff
@@ -360,7 +445,39 @@ def _async_submit_note(self, sg, data):
...
+        Uploads any generic file attachments to Shotgun, parenting
+        them to the Note entity.
+
+        :param parent_entity:   The Note entity to attach the files to in SG.
+        :param sg:              A Shotgun API handle.
+        :param data:            The data dict containing an "attachments" key
+                                housing a list of file paths to attach.
+        """
+        for file_path in data.get("attachments", []):
+            if os.path.exists(file_path):
```
robblau: Should there at least be a warning if a file we are expecting to upload isn't there?

python/note_input_widget/widget.py
```diff
@@ -360,7 +445,39 @@ def _async_submit_note(self, sg, data):
...
+
+        :param file_path:       The path to the file to upload to SG.
+        :param parent_entity:   The Note entity to attach the files to in SG.
+        :param sg:              A Shotgun API handle.
+        """
+        self._bundle.log_debug(
+            "Uploading attachments (%s bytes)..." % os.path.getsize(file_path)
+        )
+        sg.upload(parent_entity["type"], parent_entity["id"], str(file_path))
+        self._bundle.log_debug("Upload complete!")
```
robblau: Is this something that happens in the foreground?  Would it be better to run it in a bg task runner?  Ideally we would be adding some decorations to the attachments so it is clear which have been uploaded and which have yet to be uploaded.

python/screen_grab/screen_grab.py
```diff
@@ -15,6 +15,10 @@
 from sgtk.platform.qt import QtCore, QtGui
 import sgtk
 
+# If set to a callable, it will be used when performing a
+# screen grab in place of the default behavior defined in
+# this module.
+SCREEN_GRAB_CALLBACK = None
```
robblau: Any way this can be associated with the class rather than being a global?

python/activity_stream/activity_stream.py
thebeeland: Looks like Sphinx pulls property docs from the method specified as the property's getter.

python/note_input_widget/widget.py
thebeeland: Added a log_warning() if the file doesn't exist.

python/note_input_widget/widget.py
thebeeland: The entire note submission, including the uploading of any attachments, is running in the background.

PR #19, opened 2016-05-07
#36162 Note/Reply entity creation is broadcast via an entity_created signal.

thebeeland:
This allows higher-level widgets to perform additional tasks or alter the resulting entity after it is created.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -32,6 +32,9 @@ class ActivityStreamWidget(QtGui.QWidget):
     :signal playback_requested(dict): Fires when someone clicks the playback url
             on a version. Returns a shotgun dictionary with information
             about the version.
+    :signal entity_created(object): Fires when a Note or Reply entity is created by
+            an underlying widget within the activity stream. Returns a Shotgun dictionary
+            with information about the new Entity.
     :ivar reply_dialog: When a ReplyDialog is active it can be accessed here. If there
                         is no ReplyDialog active, then this will be set to None.
     :vartype reply_dialog: .dialog_reply.ReplyDialog or None
@@ -39,9 +42,15 @@ class ActivityStreamWidget(QtGui.QWidget):
     # max number of items to show in the activity stream.
     MAX_STREAM_LENGTH = 20
     entity_requested = QtCore.Signal(str, int)
-    playback_requested = QtCore.Signal(dict)   
-     
-    
+    playback_requested = QtCore.Signal(dict)
+
+    # Emitted when a Note or Reply entity is created. The
+    # entity type as a string and id as an int will be
+    # provided.
+    #
+    # dict(entity_type="Note", id=1234)
+    entity_created = QtCore.Signal(object)
+
     def __init__(self, parent):
         """
         :param parent: QT parent object
@@ -82,6 +91,7 @@ class ActivityStreamWidget(QtGui.QWidget):
         self._data_manager.update_arrived.connect(self._process_new_data)
         self._data_manager.thumbnail_arrived.connect(self._process_thumbnail)
         self.ui.note_widget.data_updated.connect(self._on_note_submitted)
+        self.ui.note_widget.entity_created.connect(self._on_entity_created)
         
         # keep handles to all widgets to be nice to the GC
         self._loading_widget = None
@@ -94,12 +104,17 @@ class ActivityStreamWidget(QtGui.QWidget):
         self._entity_type = None
         self._entity_id = None
 
+        # We'll be keeping a persistent reply dialog available because
+        # we need to connect to a signal that it's emitting. It's easiest
+        # to do that if we're dealing with an object that persists.
         self.reply_dialog = ReplyDialog(
             self,
             self._task_manager,
             note_id=None,
             allow_screenshots=self._allow_screenshots,
         )
+
+        self.reply_dialog.note_widget.entity_created.connect(self._on_entity_created)
 
     def set_bg_task_manager(self, task_manager):
         """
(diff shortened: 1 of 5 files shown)


python/note_input_widget/widget.py
```diff
@@ -42,6 +42,11 @@ class NoteInputWidget(QtGui.QWidget):
     # emitted when shotgun has been updated
     data_updated = QtCore.Signal()
     close_clicked = QtCore.Signal()
+
+    # Emitted when a Note or Reply entity is created. The
+    # entity type as a string and id as an int will be
+    # provided.
+    entity_created = QtCore.Signal(object)
```
robblau: Should provide an example of what the argument will look like in the comment.

python/activity_stream/dialog_reply.py
```diff
@@ -17,7 +17,7 @@ class ReplyDialog(QtGui.QDialog):
     This is used when someone clicks on the reply button for a note.
     """
     
-    def __init__(self, parent, bg_task_manager, note_id, allow_screenshots=True):
+    def __init__(self, parent, bg_task_manager, note_id=None, allow_screenshots=True):
```
robblau: Should add note_id to the docstring.

python/note_input_widget/widget.py
manneohrstrom: Please update the class docstring to reflect this new signal in the docs.

python/activity_stream/activity_stream.py
```diff
@@ -126,8 +130,7 @@ def note_widget(self):
         Returns the NoteInputWidget contained within the ActivityStreamWidget.
         Note that this is the widget used for NEW note input and not Note
         replies. To get the NoteInputWidget used for Note replies, access can
-        be found via reply_widget.note_widget if a reply widget is currently
-        active.
+        be found via reply_dialog.note_widget.
```
manneohrstrom: Please make the reference a link, e.g. `:meth:`ReplyDialog.note_widget``

python/activity_stream/activity_stream.py
```diff
@@ -94,8 +94,12 @@ def __init__(self, parent):
         self._entity_type = None
         self._entity_id = None
 
-        # When a ReplyDialog is active it will be stored here.
-        self.reply_dialog = None
+        self.reply_dialog = ReplyDialog(
```
manneohrstrom: Can we please add a comment here explaining to future code readers that this reply dialog is popped up and positioned whenever a reply ui is needed (by _on_reply_clicked).

Out of interest, what is the benefit of having a single dialog instance that is being reused over making it every time? Now the code feels less contained. 

Heads up that I tried initially building this as inline UI but had issues with pyside crashing, this is why I ended up with this wonky implementation where we pop up a UI and position it on top of the existing UI. Would be great if at some point we would be able to get rid of this dialog popping up altogether!

python/activity_stream/dialog_reply.py
```diff
@@ -32,14 +32,44 @@ def __init__(self, parent, bg_task_manager, note_id, allow_screenshots=True):
...
         self.ui.note_widget.set_current_entity("Note", note_id)
         self.ui.note_widget.allow_screenshots(allow_screenshots)
         
-        self.setWindowFlags(QtCore.Qt.Dialog | QtCore.Qt.FramelessWindowHint)        
+        self.setWindowFlags(QtCore.Qt.Dialog | QtCore.Qt.FramelessWindowHint)
+
+    @property
+    def note_widget(self):
+        """
+        Returns the underlying note_input_widget.NoteInputWidget.
```
manneohrstrom: Please make it sphinx friendly:
```
:class:`~note_input_widget.NoteInputWidget`
```

python/activity_stream/dialog_reply.py
```diff
@@ -32,14 +32,44 @@ def __init__(self, parent, bg_task_manager, note_id, allow_screenshots=True):
...
+        Gets the entity id number for the parent Note entity
+        being replied to.
+
+        :returns:   int
+        """
+        return self._note_id
+
+    def _set_note_id(self, note_id):
+        """
+        Sets the entity id number for the parent Note entity
```
manneohrstrom: FYI In sphinx, the getter is used as the docstring to describe the read/write property, so better to make it a generic doc string to describe the property, e.g. 

```
def _set_note_id(self, note_id):
    """
    The entity id number for the parent Note entity being replied to.
    """
```

python/activity_stream/widget_note.py
```diff
@@ -225,10 +225,10 @@ def set_note_info(self, data):
         self._set_timestamp(data, self.ui.date)
         
         # set the main note text
-        self.ui.content.setText(data["content"])
+        self.ui.content.setText(data.get("content", ""))
```
manneohrstrom: Should this be `self.ui.content.setText(data["content"] or None)` ? (Shotgun can return None for an empty field.)

To me, it feels like we should ideally be very clear about which fields `data` is assumed to contain (and updating my terrible docstring for `set_note_info` - sorry!!!!) rather than patching.

If we do expect missing key from the dict, it would be awesome if we could mention it in the docstring.

Sorry for being picky and thanks for cleaning this stuff up.

python/note_input_widget/widget.py
```diff
@@ -444,7 +450,8 @@ def _async_submit_note(self, sg, data):
         
         self.__upload_thumbnail(sg_note_data, sg, data)
         self.__upload_attachments(sg_note_data, sg, data)
-        
+
+        return sg_note_data
```
manneohrstrom: if you could update the docstring for this method while you are changing the return value, that would be amazing.

python/note_input_widget/widget.py
```diff
@@ -306,8 +311,9 @@ def _async_submit_reply(self, sg, data):
                       {"addressings_cc": updated_links})
             
         self.__upload_thumbnail(note_link, sg, data)
-        self.__upload_attachments(note_link, sg, data)     
-                
+        self.__upload_attachments(note_link, sg, data)
+
+        return sg_reply_data
```
manneohrstrom: please update docstring

python/activity_stream/dialog_reply.py
manneohrstrom: Sorry, doh, meant to add this comment to the `_get_note_id` method, not the `_set_note_id` method.

manneohrstrom:
Added some additional comments although I am not assigned the CR.

python/activity_stream/activity_stream.py
thebeeland: I need to connect to a signal, so I need a dialog that persists.

python/activity_stream/widget_note.py
thebeeland: I'll document it, but when we have a malformed Cut item, it's possible that we end up with a missing key here. It should be polished out above here, as well, but we need to make sure we don't have KeyErrors end up with the user.

PR #20, opened 2016-05-09
PlaybackLabel can now be set to be non-interactive.

thebeeland:
If not interactive, the play icon overlay will not be drawn, click events will not trigger playback signals, and the mouse icon won't be set to a finger pointer.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -71,6 +71,7 @@ class ActivityStreamWidget(QtGui.QWidget):
         # customizations
         self._allow_screenshots = True
         self._show_sg_stream_button = True
+        self._version_items_playable = True
         
         # apply styling
         self._load_stylesheet()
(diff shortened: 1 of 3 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -185,6 +186,24 @@ def _set_show_sg_stream_button(self, state):
...
+        return self._version_items_playable
+
+    def _set_version_items_playable(self, state):
+        self._version_items_playable = bool(state)
+
+    version_items_playable = QtCore.Property(
+        bool,
+        _get_version_items_playable,
+        _set_version_items_playable,
+    )
```
robblau: Does this render well for Sphinx docs?

python/playback_label/playback_label.py
```diff
@@ -70,25 +71,46 @@ def set_shotgun_data(self, sg_data):
...
+        """
+        return self._interactive
+
+    def _set_interactive(self, state):
+        self._interactive = bool(state)
+
+        if not self._interactive:
+            self.unsetCursor()
+        elif self.playable:
+            self.setCursor(QtCore.Qt.PointingHandCursor)
```
robblau: If interactive and not playable this'll leave the cursor as is.  Is that the intended behavior?

python/activity_stream/activity_stream.py
thebeeland: The getter is what's sourced by Sphinx, so that's where the docs should go for properties across the board. If we want additional docs for the setter for the sake of a dev reading the code that's fine, but it won't show up in Sphinx docs.

PR #21, opened 2016-05-14
Exposes the ability to build field menus from a project-specific schema.

thebeeland:
This allows for project-level fields on entities to be included in field menus.


thebeeland:
This has already been incorporated into master.

PR #22, opened 2016-05-17
34772 Adds editing functionality to shotgun field widgets

josh-t:
These changes build on the work previously done by @robblau to introduce editing capabilities to all the SG field types. 

field_widgets

The changes include:

- editor widgets for all field types where editing is supported by the API (url field still in progress)
- introduction of bubble widget base classes for use by multi-entity editor and tag editor (for whenever tag editing becomes supported by the API)
- simplification of the manager class to accept a `widget_type` argument to methods rather than having separate method for each type (extra eyes would be good here to see if this feels like the right interface)
- a new class that wraps the display/editor widgets for a field into an `EDITABLE` type that allows users to switch between display and editing
- a matching class that wraps the display widget of a read only field and shows an indicator that editing is not available for that field
- sphinx docs for the field widgets and related classes with lots of screenshots
- separation of the completer from the `global_search_widget` into a reusable `global_search_completer`. (widget API remains unchanged)
- global search widget/delegate moved under completer module (they weren't exposed via API)
- completer still needs some documentation work
- elided label changes to make it work in grid/form layouts (needs extra scrutiny)


python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,349 @@
...
+
+shotgun_data = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_data")
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+
+class GlobalSearchCompleter(QtGui.QCompleter):
+
+
+    # emitted when shotgun has been updated
+    entity_selected = QtCore.Signal(str, int)
+    entity_activated = QtCore.Signal(str, int, str)
```
robblau: The arguments to the signals should be documented to make them easier to use.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,343 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk.platform.qt import QtCore, QtGui
+
+
+class BubbleEditWidget(QtGui.QTextEdit):
+
+    OBJECT_REPLACEMENT_CHAR = unichr(0xfffc)
```
robblau: Should have class docstring and comments for the constant.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,343 @@
...
+        for format in self._formats.values():
+            del format
+        self._bubble_text_object.clear()
+        super(BubbleEditWidget, self).clear()
+
+    def add_bubble(self, bubble):
+
+        bubble_id = self._bubble_text_object.add_bubble(bubble)
+
+        format = QtGui.QTextCharFormat()
```
robblau: Code is good, but some comments would be good.

python/shotgun_fields/date_widget.py
```diff
@@ -30,5 +55,70 @@ def _string_value(self, value):
...
+            return value.toPython()
+
+    def keyPressEvent(self, event):
+        """
+        Provides shortcuts for applying modified values.
+
+        :param event: The key press event object
+        :type event: :class:`~PySide.QtGui.QKeyEvent`
+        """
+        if event.key() in [QtCore.Qt.Key_Enter, QtCore.Qt.Key_Return]:
```
robblau: This keeps showing up.  Is there somewhere in the meta class we could put logic like this?

python/shotgun_fields/file_link_widget.py
```diff
@@ -8,20 +8,130 @@
...
+        self._popup_btn = QtGui.QPushButton(self)
+        self._popup_btn.setIcon(QtGui.QIcon(":/qtwidgets-shotgun-fields/link_menu.png"))
+        self._popup_btn.setFixedSize(QtCore.QSize(18, 12))
+        self._popup_btn.setFocusPolicy(QtCore.Qt.NoFocus)
+        self._popup_btn.hide()
+
+        # make sure there's never a bg color or border
+        self._popup_btn.setStyleSheet("background-color: none; border: none;")
+
+        # XXX implement all the actions
```
robblau: This still needed?

python/shotgun_fields/file_link_widget.py
```diff
@@ -30,11 +140,44 @@ def _string_value(self, value):
...
+            self,
+            caption="Replace Image",
+            options=QtGui.QFileDialog.DontResolveSymlinks,
+        )[0]
+        if file_path and os.path.exists(file_path):
+            self._clear()
+
+            # populte dict to match SG value
+            self.set_value({
+                # XXX
```
robblau: TBD, right?

python/global_search_completer/global_search_completer.py
josh-t: Done.

python/shotgun_fields/date_widget.py
josh-t: These particular "apply" key shortcuts show up in 6 or so of the line edit based widgets. Other, text edit based widgets, use different keys. Other widgets don't use key events to apply values. To me it feels like there's not enough similarity across all the widgets to put something in the metaclass. I'm happy to though if you have something you think is reasonable.

python/elided_label/elided_label.py
```diff
@@ -69,6 +102,10 @@ def setText(self, text):
 
         :param text:    The text to set on the label
         """
+        # clear out the ideal width so that the widget can recalculate based on
+        # the new text
+        if hasattr(self, '_ideal_width'):
```
thebeeland: Minor: would it be cleaner for it to always exist but contain None when it's unpopulated?

python/elided_label/elided_label.py
```diff
@@ -141,20 +184,30 @@ def _elide_text(self, text, elide_mode):
...
+                    break
+
+            self._line_width = line_width
             return doc.toHtml()
         finally:
             # clean up the doc:
             doc.deleteLater()
 
+    @property
+    def line_width(self):
```
thebeeland: docstring

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
+# Copyright (c) 2015 Shotgun Software Inc.
```
thebeeland: Minor: 2016 vs. 2015

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        """
+        Should be called before the widget is closed
+        """
+        if self.__sg_data_retriever:
+            self.__sg_data_retriever.stop()
+            self.__sg_data_retriever.work_completed.disconnect(self.__on_worker_signal)
+            self.__sg_data_retriever.work_failure.disconnect(self.__on_worker_failure)
+            self.__sg_data_retriever = None
+
+    def get_current_result(self):
```
thebeeland: docstring

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        model_index = self.popup().currentIndex()
+        return self.get_result(model_index)
+
+    def get_result(self, model_index):
```
thebeeland: docstring

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        popup = self.popup()
+        self._delegate = SearchResultDelegate(popup)
+        popup.setItemDelegate(self._delegate)
+
+        # try to disconnect and reconnect the activated signal
+        # it seems this signal is lost every time the widget
+        # looses focus.
+        try:
+            self.activated[QtCore.QModelIndex].disconnect(self._on_select)
+        except:
```
thebeeland: use `except Exception:` instead

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+            )
+
+        self.activated[QtCore.QModelIndex].connect(self._on_select)
+
+        # now clear the model
+        self._clear_model()
+
+        # clear download queue
+        if self.__sg_data_retriever:
+            self.__sg_data_retriever.clear()
```
thebeeland: You can do this below inside the conditional at 187

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        # thread when the worker queue reaches that point and will
+        # call out to execute it. The data dictionary specified will
+        # be forwarded to the method.
+        data = {"text": text}
+        if self.__sg_data_retriever:
+            self._processing_id = self.__sg_data_retriever.execute_method(
+                self._do_sg_global_search, data)
+        else:
+            raise sgtk.TankError(
+                "Please associate this class with a background task processor.")
```
thebeeland: You mean Shotgun Data Retriever?

thebeeland: Ah, perhaps not now that I've looked at set_bg_task_manager. Perhaps just replace "processor" with "manager"

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        :param sg: Shotgun instance
+        :param data: data dictionary passed in from _submit(). This dict
+            carries the options payload and is forwarded over by the
+            sg_data.execute_method() method. In our case, the dictionary
+            holds a single 'text' key which contains the search phrase.
+        """
+
+        # constrain by project in the search
+        project_ids = []
+        if self._bundle.context.project:
```
thebeeland: I'm curious whether Rob/Manne prefer this or the following:
```
self._bundle.tank.pipeline_configuration.get_project_id()
```

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+        Fires when an item in the completer is selected.
+        This will emit an entity_selected signal for the
+        global search widget
+
+        :param model_index: QModelIndex describing the current item
+        """
+        data = self.get_result(model_index)
+        if data:
+            self.entity_selected.emit(data["type"], data["id"])
+            self.entity_activated.emit(data["type"], data["id"], data["name"])
```
thebeeland: Is there ever a case when one of these is emitted but not the other? If the answer is that there isn't, does it make sense to have both?

python/global_search_widget/global_search_widget.py
```diff
@@ -8,67 +8,41 @@
...
-        self._thumb_map = {}
-        self._default_icon = QtGui.QPixmap(":/tk_framework_qtwidgets.global_search_widget/rect_512x400.png")
-                
         # configure our popup completer
-        self._completer = QtGui.QCompleter(self)
-        self._completer.setMaxVisibleItems(10)
-        self._completer.setCaseSensitivity(QtCore.Qt.CaseInsensitive)
-        self._completer.setCompletionMode(QtGui.QCompleter.UnfilteredPopupCompletion)
-        self.setCompleter(self._completer)
+        self.setCompleter(global_search_copmleter.GlobalSearchCompleter(self))
```
thebeeland: typo: copmleter

python/global_search_completer/global_search_completer.py
```diff
@@ -0,0 +1,362 @@
...
+            for d in matches:
+                item = QtGui.QStandardItem(d["name"])
+                item.setData(self.MODE_RESULT, self.MODE_ROLE)
+
+                item.setData(shotgun_model.sanitize_for_qt_model(d),
+                             self.SG_DATA_ROLE)
+
+                item.setIcon(self._default_icon)
+
+                if d.get("image", None) and self.__sg_data_retriever:
```
thebeeland: Is it possible not have __sg_data_retriever given that we're in the work completed callback here? Seems like it pretty much has to be available at this point.

thebeeland: Also, you can just do `d.get("image")` since the default for the default is None.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk.platform.qt import QtCore, QtGui
+
+
+class BubbleWidget(QtGui.QFrame):
+    """
+    This class represents a drawable "bubble" ot display in a :class:`.BubbleEditWidget`
```
thebeeland: typo: ot

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        super(BubbleWidget, self).__init__(parent)
+
+        # placeholder for the underlying data this widget represents in the editor
+        self._data = None
+
+        # should not grow or shrink
+        self.setSizePolicy(QtGui.QSizePolicy.Fixed, QtGui.QSizePolicy.Fixed)
+
+        # style the look of the bubble
+        self.setObjectName("bubble")
```
thebeeland: What happens if a bunch of these widgets end up with the same object name? I'm not sure this is necessary since the qss is being applied directly to the widget. That should mean you can scope that qss to BubbleWidget and it won't leak into anything else.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        Shows or hides the ``x`` button depending on the value of the
+        ``removable`` argument.
+        :param bool removable: ``True`` if the bubble is removable, ``False`` otherwise.
+        """
+        if removable:
+            self.remove_button.show()
+        else:
+            self.remove_button.hide()
+
+    def get_data(self):
```
thebeeland: I prefer a property or just an instance variable rather than getter and setter methods, but it could also be argued that this better aligns with Qt's style. It might be worth us talking about this as a group so that as we continue to build qtwidgets that we're consistent.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+
+        :param bubble: The bubble widget instance.
+        :return: A unique id for the added bubble
+        :rtype: :obj:`int`
+        """
+
+        bubble_id = self._bubble_text_object.add_bubble(bubble)
+
+        # create a character format for the bubble
+        format = QtGui.QTextCharFormat()
```
thebeeland: `format` is a built-in function, so I would suggest a different variable name here

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        """
+        for format in self._formats.values():
+            del format
+        self._bubble_text_object.clear()
+        super(BubbleEditWidget, self).clear()
+
+    def clear_typed_text(self):
+        """
+        Clears only typed text (not bubbles) from the editor.
+        :return:
```
thebeeland: Looks like an incomplete docstring.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        """
+
+        text = self.toPlainText()
+        cursor = self.textCursor()
+
+        cursor.beginEditBlock()
+
+        # iterate over all characters in the editor. remove any that aren't the
+        # special replacement character. iterate backwards so that indices don't change
+        for i in reversed(range(0, len(text))):
```
thebeeland: `xrange(len(text), 0, -1)`

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        # resulting to just adding the scroll values. please fix if you can
+        edit_pos = QtCore.QPoint(
+            event.pos().x() + self.horizontalScrollBar().value(),
+            event.pos().y() + self.verticalScrollBar().value()
+        )
+
+        # for mouse events find the actual widget at the position
+        doc = self.document()
+        cursor_pos = doc.documentLayout().hitTest(edit_pos, QtCore.Qt.ExactHit)
+        format = doc.documentLayout().format(cursor_pos)
```
thebeeland: Same here as before. I'd try to stay away from variable names that match builtins.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+
+        if not bubble_id in self._formats:
+            return None
+
+        text = self.toPlainText()
+        cursor = self.textCursor()
+
+        # loop over each character until a replacement character with a known
+        # format that matches the supplied id is found.
+        for i in range(0, len(text)):
```
thebeeland: Minor: `xrange` is technically more memory efficient since it's a generator and the length of the text here isn't known.

python/shotgun_fields/bubble_widget.py
thebeeland: FYI, I won't mention this again, but I see others.

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+        if not self.get_bubble(bubble_id):
+            return
+
+        text = self.toPlainText()
+        cursor = self.textCursor()
+
+        # locate the bubble by iterating over each character in the editor.
+        # when the match is found, remove it.
+        for i in range(0, len(text)):
+            if not text[i] == self._OBJECT_REPLACEMENT_CHAR:
```
thebeeland: Minor: `!=` rather than `not ... ==`

python/shotgun_fields/bubble_widget.py
```diff
@@ -0,0 +1,495 @@
...
+            # we found the bubble's object character.
+            cursor.beginEditBlock()
+            cursor.movePosition(QtGui.QTextCursor.Left, QtGui.QTextCursor.KeepAnchor)
+            cursor.removeSelectedText()
+            cursor.endEditBlock()
+
+            del self._formats[bubble_id]
+
+            self.update()
+            return
```
thebeeland: The `return` isn't required.

python/shotgun_fields/currency_widget.py
```diff
@@ -8,27 +8,93 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
-"""
-Widget that represents the value of a currency field in Shotgun
-"""
 import locale
+from sgtk.platform.qt import QtGui, QtCore
```
thebeeland: I don't immediately see QtGui being used.

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -68,3 +53,82 @@ def _create_human_readable_timestamp(self, dt, postfix=""):
...
+    __metaclass__ = ShotgunFieldMeta
+    _EDITOR_TYPE = "date_time"
+
+    def get_value(self):
+        """
+        :return: The internal value being displayed by the widget.
+        """
+        value = self.dateTime()
+
+        if hasattr(QtCore, "QVariant"):
```
thebeeland: By far the majority case here should end up being PySide, so why not just put that in a try and except on AttributeError to try the PyQt method?

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -68,3 +53,82 @@ def _create_human_readable_timestamp(self, dt, postfix=""):
...
+            super(DateAndTimeEditorWidget, self).keyPressEvent(event)
+
+    def setup_widget(self):
+        """
+        Prepare the widget for display.
+
+        Called by the metaclass during initialization.
+        """
+        self.setCalendarPopup(True)
+        self.setMinimumWidth(100)
```
thebeeland: I'd prefer the 100 be assigned to a class variable and then that being referenced here.

python/shotgun_fields/date_widget.py
```diff
@@ -8,20 +8,45 @@
...
     """
-    Inherited from a :class:`~DateAndTimeWidget`, this class is able to
-    display a date field value as returned by the Shotgun API.
+    Display a ``date`` field value as returned by the Shotgun API.
     """
-    _FIELD_TYPE = "date"
+    __metaclass__ = ShotgunFieldMeta
+    _DISPLAY_TYPE = "date"
+
+    def _create_human_readable_date(self, date):
```
thebeeland: FYI, this logic has been moved into shotgun_globals (I think that's where I put it...) in newer code.

thebeeland: The above is likely going to be shown when you go to resolve merge conflicts, because I did that after you started your work here.

python/global_search_completer/global_search_completer.py
manneohrstrom: Conceptually, Context represents a location in Shotgun, whereas pipeline_configuration.get_project_id() represents the project that the current project is associated with. In the case of any shotgun related stuff like this, use context. 

I think right now, they will always be the same, but if we wanted to allow for more cross project action going on in the future, we would be better set up with this being context based (as it is now).

python/shotgun_fields/date_and_time_widget.py
```diff
@@ -68,3 +53,82 @@ def _create_human_readable_timestamp(self, dt, postfix=""):
...
+
+        :param value: The value to convert into a string
+        :type value: :class:`datetime.datetime` or a float representing unix time
+        """
+
+        # shotgun_model converts datetimes to floats representing unix time so
+        # handle that as a valid value as well
+        if not isinstance(value, datetime.datetime):
+            value = datetime.datetime.fromtimestamp(value)
+        return self._create_human_readable_timestamp(value, " %I:%M%p")
```
manneohrstrom: comment with example would be great here explaining what `%I:%M%p` means.

manneohrstrom: is it possible / does it make sense to have a tooltip that shows the exact value. I think this is how the activity stream works and i like how in github it says "you committed this _10 days ago_" and if you hover over the date you see the exact timestamp.

python/elided_label/elided_label.py
josh-t: Changed.

python/elided_label/elided_label.py
josh-t: added.

python/global_search_completer/global_search_completer.py
josh-t: fixed.

python/global_search_completer/global_search_completer.py
josh-t: added.

python/global_search_completer/global_search_completer.py
josh-t: changed.

python/global_search_completer/global_search_completer.py
josh-t: moved.

python/global_search_completer/global_search_completer.py
josh-t: Much of this is copy/paste form removing the completer from the widget. Yeah, I think "manager" is correct here. changed.

python/global_search_completer/global_search_completer.py
josh-t: Good to know.

python/global_search_completer/global_search_completer.py
josh-t: don't know about the __sg_data_retriever. seems safe to leave for now since it's being used right after. i've changed the `get()` to remove `None`.

python/global_search_completer/global_search_completer.py
josh-t: The one with 2 args is the one that existed before in the widget. I kept it for backward compatibility (see the docs way up at the top). After a chat with @manneohrstrom I added the second signal with the additional arg to make the returned data easier to display without doing another round trip to SG if it wasn't absolutely necessary.

python/global_search_widget/global_search_widget.py
josh-t: wow! spelled it wrong on the import too. double whammy. well, it was working!!! fixed.

python/shotgun_fields/bubble_widget.py
josh-t: fixed.

python/shotgun_fields/bubble_widget.py
josh-t: I'll defer to @robblau as I think he added this bit. perhaps he tried setting stylesheet without limiting it to a named widget? i will see if it works without naming, but don't think it will hurt to leave as is... will run some tests.

python/shotgun_fields/bubble_widget.py
josh-t: The idea here was to be a little closer to how Qt handles get/set of data for widgets. This also aligns with the fields widgets interface of get/set value. Feel free to add to hot topics if you think we should take a different approach.

python/shotgun_fields/bubble_widget.py
josh-t: good call. changed all `format` variations to `char_format`. need to go back and test to make sure i didn't miss anything.

python/shotgun_fields/bubble_widget.py
josh-t: fixed.

python/shotgun_fields/bubble_widget.py
josh-t: changed.

python/shotgun_fields/bubble_widget.py
josh-t: cool. yep, all changed.

python/shotgun_fields/bubble_widget.py
josh-t: changed.

python/shotgun_fields/bubble_widget.py
josh-t: good call. found a bunch of others in here too. i think i did it this way once and cut/pasted a bunch.

python/shotgun_fields/bubble_widget.py
josh-t: actually it's inside the `for` loop as its expected that there's only one matching bubble, the method returns after it removes the one.

python/shotgun_fields/currency_widget.py
josh-t: used below:

`class CurrencyEditorWidget(QtGui.QDoubleSpinBox)`

python/shotgun_fields/date_and_time_widget.py
josh-t: Added additional docs for the timestamp formatting. will add tooltips as well.

python/shotgun_fields/date_and_time_widget.py
josh-t: changed.

python/shotgun_fields/date_and_time_widget.py
josh-t: done.

python/shotgun_fields/date_widget.py
josh-t: sweet! thanks for the heads up.

python/shotgun_fields/footage_widget.py
```diff
@@ -8,17 +8,148 @@
...
+        """
+        return str(value)
+
+
+class _FootageInputValidator(QtGui.QValidator):
+    """
+    A validator for the {feet}-{frames} footage spec.
+    """
+
+    def fixup(self, input):
```
josh-t: another good catch. changed to `input_str` for all cases.

python/shotgun_fields/footage_widget.py
```diff
@@ -8,17 +8,148 @@
...
+            string.
+        :rtype tuple:
+
+        :raises: ``ValueError`` if feet and frames cannot be inferred.
+        """
+
+        input = input.strip()
+        input = input.rstrip("-")
+
+        match = re.match("^\d+$", input)
```
josh-t: cool. changed.

python/shotgun_fields/image_widget.py
```diff
@@ -8,124 +8,320 @@
...
+        """
+        Display a file browser and process the selected file path.
+        """
+
+        file_path = QtGui.QFileDialog.getOpenFileName(
+            self,
+            caption="Replace Image",
+            options=QtGui.QFileDialog.DontResolveSymlinks,
+        )[0]
+        if file_path and os.path.exists(file_path):
```
josh-t: nope. removed.

python/shotgun_fields/list_widget.py
```diff
@@ -8,17 +8,73 @@
...
+
+        Called by the metaclass during initialization. Adds the valid values to
+        the list and connects the ``activated`` signal.
+        """
+        self.addItem("")
+        valid_values = shotgun_globals.get_valid_values(self._entity_type, self._field_name)
+        self.addItems(valid_values)
+
+        self.activated.connect(
+            lambda i: self.value_changed.emit()
```
josh-t: the `activated` signal sends an index argument which we want to ignore since emit does not accept an argument. we can get at it via the `currentIndex` method later.

python/shotgun_fields/percent_widget.py
```diff
@@ -8,27 +8,81 @@
...
+    __metaclass__ = ShotgunFieldMeta
+    _EDITOR_TYPE = "percent"
+
+    def get_value(self):
+        """
+        :return: The internal value being displayed by the widget.
+        """
+        return self.value()
+
+    def keyPressEvent(self, event):
```
josh-t: rob asked the same...

> These particular "apply" key shortcuts show up in 6 or so of the line edit based widgets. Other, text edit based widgets, use different keys. Other widgets don't use key events to apply values. To me it feels like there's not enough similarity across all the widgets to put something in the metaclass. I'm happy to though if you have something you think is reasonable.

Thinking more about it, we could add an intermediate `QLineEdit` base class that has this logic. But then there are the date/datetime widgets that have their own internal `QLineEdit`. I'm inclined to leave it for now since the code is pretty minimal. Let me know if you think otherwise.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -20,20 +20,30 @@ class ShotgunFieldDelegate(views.WidgetDelegate):
...
+        :type field_name: String
+
+        :param display_class: A shotgun field :class:`~PySide.QtGui.QWidget` to display the field info
+
+        :param editor_class: A shotgun field :class:`~PySide.QtGui.QWidget` to edit the field info
 
         :param view: The parent view for this delegate
         :type view:  :class:`~PySide.QtGui.QWidget`
         """
         views.WidgetDelegate.__init__(self, view)
```
josh-t: I've made a note to go back through and update these.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -43,8 +53,13 @@ def _create_widget(self, parent):
...
         :rtype:         :class:`~PySide.QtGui.QWidget`
         """
-        self._display_widget.setParent(parent)
-        return self._display_widget
+        widget = self._display_class(
+            entity_type=self._entity_type,
+            field_name=self._field_name,
+            bg_task_manager=self._bg_task_manager,
+        )
+        widget.setParent(parent)
```
josh-t: Yep, I need to go back in and clean up the delegate class. I haven't really touched it since early on.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -81,6 +103,16 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+        self.setEditorData(widget, model_index)
+
+    def setEditorData(self, widget, index):
+        value = index.data(shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE)
         sanitized_value = shotgun_model.sanitize_qt(value)
-        self._display_widget.set_value(sanitized_value)
+        widget.set_value(sanitized_value)
+
+    def setModelData(self, editor, model, index):
+        print "setModelData: %s" % editor.get_value()
```
josh-t: yep. still WIP. not touched much.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -81,6 +103,16 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+
+    def editorEvent(self, event, model, option, index):
+        # print "EDITOR EVENT: %s" % event
```
josh-t: same.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+        # TODO: insert backend update here (don't immediately apply the value)
+        self._editor.edit_widget.value_changed.connect(self._apply_value)
+
+        self.currentChanged.connect(self._on_current_changed)
+
+    def minimumSizeHint(self):
+        """
+        Returns the minimum size hint for the currently displayed widget
+        """
+        return self.currentWidget().minimumSizeHint()
```
josh-t: the widgets get set in `__init__` so I think it should be ok.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def sizeHint(self):
+        """
+        Returns the size hint for the currently displayed widget
+        """
+        return self.currentWidget().sizeHint()
```
josh-t: the widgets get set in `__init__` so I think it should be ok.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+
+    def _on_current_changed(self, index):
+        """
+        Primarily used to ensure focus and to make sure the display/edit widgets are in sync.
+
+        :param int index: The index of the newly current widget in the stack.
+        """
+
+        if index == self._edit_index:
+            self._editor.edit_widget.blockSignals(True)
```
josh-t: switched to `try/finally`

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+
+        layout = QtGui.QHBoxLayout(self)
+        layout.setContentsMargins(0, 0, 0, 0)
+        layout.setSpacing(0)
+        layout.addWidget(display_widget)
+        layout.addWidget(spacer)
+        layout.addWidget(self._edit_btn)
+        layout.addStretch(10)
+
+        self.setMinimumHeight(self._edit_btn.height())
```
josh-t: Because the button only shows on hover, this is required. Otherwise things get jumpy.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+            btn_layout.addWidget(self._done_btn)
+            btn_layout.addStretch()
+            btn_layout.addWidget(self._apply_btn)
+        else:
+            btn_layout = QtGui.QHBoxLayout()
+            btn_layout.addWidget(self._apply_btn)
+            btn_layout.addWidget(self._done_btn)
+            btn_layout.addStretch()
+
+        if hasattr(editor_widget, '_IMMEDIATE_APPLY') and editor_widget._IMMEDIATE_APPLY:
```
josh-t: updated.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -0,0 +1,320 @@
...
+        layout = QtGui.QHBoxLayout(self)
+        layout.setContentsMargins(0, 0, 0, 0)
+        layout.setSpacing(4)
+        layout.addWidget(editor_widget)
+        layout.addLayout(btn_layout)
+        layout.addStretch()
+
+        layout.setAlignment(self._done_btn, QtCore.Qt.AlignBottom)
+
+        #self.setMinimumHeight(self._done_btn.height())
```
josh-t: removed. not needed since it's always visible when the editor widget is shown.

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -107,39 +22,101 @@ class ShotgunFieldManager(QtCore.QObject):
...
     initialized = QtCore.Signal()
 
+    # widget types enumeration
+    _WIDGET_TYPES = (DISPLAY, EDITOR, EDITABLE) = ("display", "editor", "editable")
+
+    ############################################################################
+    # class methods
+
+    @classmethod
+    def get_class(cls, sg_entity_type, field_name, widget_type=DISPLAY):
```
josh-t: i thought that initially too. but this method is part of the `ShotgunFieldManager` definition so that namespace isn't available yet. but the widget type attributes have been defined (they've above this line) so this seems to work. Not sure if this is good practice though. not sure I can remember doing this before.

python/shotgun_fields/shotgun_field_manager.py
```diff
@@ -151,17 +128,98 @@ def __init__(self, parent, bg_task_manager=None):
...
+        :returns: :class:`~PySide.QtGui.QWidget` or ``None`` if the field type has no display widget
+        """
+
+        if widget_type is self.EDITABLE:
+            widget = self._create_editable_widget(
+                sg_entity_type, field_name, entity, parent, **kwargs)
+        elif widget_type is self.EDITOR:
+            widget = self._create_editor_widget(
+                sg_entity_type, field_name, entity, parent, **kwargs)
+        else:
```
josh-t: now raises. :)

python/shotgun_fields/shotgun_field_meta.py
```diff
@@ -9,110 +9,182 @@
...
+            raise ValueError("ShotgunFieldMeta classes must have a _DISPLAY_TYPE or _EDITOR_TYPE member variable")
         if "__init__" in class_dict:
             raise ValueError("ShotgunFieldMeta classes cannot define their own constructor")
 
-        # take over interface methods if they have not already been defined
-        mcl.take_over_if_not_defined("setup_widget", mcl.setup_widget, class_dict, parents)
-        mcl.take_over_if_not_defined("set_value", mcl.set_value, class_dict, parents)
+        # take over class members if called out via the @take_over decorator
+        for member_name in TAKE_OVER_NAMES:
+            if hasattr(mcl, member_name):
```
josh-t: changed to use `getattr()` with a default. :)

python/shotgun_fields/status_list_widget.py
```diff
@@ -41,3 +37,63 @@ def _string_value(self, value):
...
+        self.setCurrentIndex(0)
+
+    def _display_value(self, value):
+        """
+        Set the value displayed by the widget.
+
+        :param value: The value returned by the Shotgun API to be displayed
+        """
+        if value is None:
+            self.setCurrentIndex(0)
```
josh-t: Yep. and tweaked the code below in case the value couldn't be found to set to 0 as well.

python/shotgun_fields/bubble_widget.py
josh-t: I played around with this a little bit. The bubble widget contains 2 labels (both subclasses of QFrame) and a button. Without isolating the style to a named widget (the top-level bubble), the style does end up being used by the internal labels. I'm inclined to leave this as-is.

python/shotgun_fields/date_and_time_widget.py
josh-t: tooltips added. :)

PR #23, opened 2016-06-07
Refactors how annotated frames are attached to Notes and Replies.

thebeeland:
This allows for a pre-submit callback to be attached to a NoteInputWidget. This can be used to do things like procedurally add attachments to the NoteInputWidget prior to the Note's or Reply's creation.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -115,6 +115,14 @@ class ActivityStreamWidget(QtGui.QWidget):
             allow_screenshots=self._allow_screenshots,
         )
 
+        # We'll allow for a pre-note-creation callback. This is for additional
+        # pre-processing that needs to occur before a Note or Reply is created
+        # in Shotgun. This makes sure that the activity stream data coming down
+        # during the rescan after submission contains anything like additional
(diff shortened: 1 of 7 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -150,6 +158,28 @@ def note_widget(self):
...
+        :returns:   Python callable or None
+        """
+        return self._pre_submit_callback
+
+    def _set_pre_submit_callback(self, callback):
+        self._pre_submit_callback = callback
+        self.reply_dialog.note_widget.pre_submit_callback = callback
+        self.note_widget.pre_submit_callback = callback
+
+    pre_submit_callback = QtCore.Property(
```
josh-t: Is `Property` specific to `PySide`?

josh-t: Also, do you know if the docs for the property get exposed when using this form?

thebeeland: Ugh...QProperties do exist in PyQt4, but they're not setup the same way.

Oh well, I'll switch them to be normal properties. The only advantage to QProperties is type validation and the ability to interface with them from Qt Designer if we were to set this widget up for drag/drop use there.

PR #24, opened 2016-06-15
#36884: Signals passed up for note arrival and entity loading.

thebeeland:
Plus access to note thread data as a property of activity stream and version details widgets.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -22,6 +22,7 @@ from .widget_value_update import ValueUpdateWidget
 from .dialog_reply import ReplyDialog
 from .data_manager import ActivityStreamDataHandler
 from .overlaywidget import SmallOverlayWidget
+from ..note_input_widget import NoteInputDialog
  
 class ActivityStreamWidget(QtGui.QWidget):
     """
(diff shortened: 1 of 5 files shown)


python/activity_stream/data_manager.py
```diff
@@ -86,6 +86,7 @@ class ActivityStreamDataHandler(QtCore.QObject):
     
     update_arrived = QtCore.Signal(list)
     note_arrived = QtCore.Signal(int, int)
+    note_thread_arrived = QtCore.Signal(int, object)
     thumbnail_arrived = QtCore.Signal(dict)
```
robblau: Signals should be documented... when they are emitted and what the arguments are.

python/version_details/version_details.py
```diff
@@ -60,9 +60,13 @@ class VersionDetailsWidget(QtGui.QWidget):
     # along.
     entity_created = QtCore.Signal(object)
 
+    # Emitted when an entity is loaded in the panel.
```
robblau: Argument should be doc'ed.

python/activity_stream/activity_stream.py
```diff
@@ -145,6 +146,14 @@ def destroy(self):
     # properties
 
     @property
+    def note_threads(self):
+        """
+        The currently loaded note threads, keyed by Note entity id and
+        containing a list of Shotgun entity dictionaries.
```
manneohrstrom: What are these note threads? Are they all the note threads that are expanded within the activity stream? Please extend the comment to give more context. Maybe also update the  class level doc string.

python/activity_stream/activity_stream.py
```diff
@@ -165,8 +174,7 @@ def _set_allow_screenshots(self, state):
         self._allow_screenshots = bool(state)
         self.ui.note_widget.allow_screenshots(self._allow_screenshots)
 
-    allow_screenshots = QtCore.Property(
-        bool,
+    allow_screenshots = property(
```
manneohrstrom: please add docstring

python/activity_stream/activity_stream.py
```diff
@@ -185,8 +193,7 @@ def _set_show_sg_stream_button(self, state):
         """
         self._show_sg_stream_button = bool(state)
 
-    show_sg_stream_button = QtCore.Property(
-        bool,
+    show_sg_stream_button = property(
```
manneohrstrom: please add docstring

PR #27, opened 2016-06-28
#35963 Allows editor widgets to be used as delegates with the shotgunmodel

josh-t:
This change fleshes out the use of field widgets as delegates with a SG model backed view. Changes include:

- editable views (such as sg tree view) can now edit SG model via field widget delegates
- delegates pull data from model to display in editors and update model properly when closed
- clicking on painted display delegate forwards click to underlying "real" widget. this allows urls to be clicked on to view entity/link fields in browser
- field widgets are now aware when they're being used as delegates. this allows them to be a little smarter about how they process data when necessary.

### DIFF
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -17,6 +17,7 @@ from .shotgun_field_meta import ShotgunFieldMeta
 from .ui import resources_rc
 
 shotgun_data = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_data")
+shotgun_globals = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_globals")
 
 
 class ImageWidget(QtGui.QLabel):
@@ -32,13 +33,23 @@ class ImageWidget(QtGui.QLabel):
     @property
     def image_url(self):
         """
-        The string url to the currently loaded image file.
+        *DEPRECATED* - Please use ``image_path`` instead.
         """
-        return self._image_url
+        return self.image_path
+
+    @property
+    def image_path(self):
+        """
+        The local path to the currently loaded image.
+        """
+        return self._image_path
 
     def clear(self):
+        """
+        Clears the widget's knowledge of an external resource.
+        """
         self._pixmap = None
-        self._image_url = None
+        self._image_path = None
         if not self._delegate:
             self._needs_download = True
         super(ImageWidget, self).clear()
@@ -111,6 +122,11 @@ class ImageWidget(QtGui.QLabel):
         :type event: :class:`~PySide.QtGui.QResizeEvent`
         """
         if self._pixmap:
+
+            # the pixmap stored internally is the full size pixmap. calling this
+            # method resizes it based on the size hint of the widget and
+            # preserving it's aspect ratio. the returned, scaled pixmap is then
+            # used for display
             scaled_pixmap = self._scale_pixmap(self._pixmap)
             self._scaled_width = scaled_pixmap.width()
             super(ImageWidget, self).setPixmap(scaled_pixmap)
@@ -128,6 +144,10 @@ class ImageWidget(QtGui.QLabel):
         """
 
         self._pixmap = pixmap
+
+        # the pixmap supplied is the full size pixmap. we'll keep a handle on
+        # that internally, but we'll display a scaled version of it based on the
+        # available space the widget has.
         scaled_pixmap = self._scale_pixmap(pixmap)
         self._scaled_width = scaled_pixmap.width()
         super(ImageWidget, self).setPixmap(scaled_pixmap)
@@ -140,13 +160,15 @@ class ImageWidget(QtGui.QLabel):
         Called by the metaclass during initialization.
         """
         self._pixmap = None
-        self._image_url = None
+        self._image_path = None
         self._editable = False
         self._scaled_width = self.width()
 
+        self.setAlignment(QtCore.Qt.AlignLeft | QtCore.Qt.AlignTop)
+
         if self._delegate:
             # in delegate mode. that means this widget is being used to display
-            # multiple entitie's image fields. so setting up a data retriever to
+            # multiple entity's image fields. so setting up a data retriever to
             # download for a specific entity is pointless.
             self._needs_download = False
         else:
@@ -243,7 +265,9 @@ class ImageWidget(QtGui.QLabel):
         """
         Set the value displayed by the widget.
 
-        :param value: The value returned by the Shotgun API to be displayed
+        :param value: The value returned by the Shotgun API to be displayed.
+            In delegate mode, this value can also be an existing ``QPixmap``
+            object.
         """
 
         if isinstance(value, QtGui.QPixmap):
@@ -251,8 +275,12 @@ class ImageWidget(QtGui.QLabel):
 
         if self._needs_download:
             # queue up the download in the background
-            entity_id = self._entity and self._entity.get("id") or None
-            entity_type = self._entity and self._entity.get("type") or None
+            entity_id = None
+            entity_type = None
+            if self._entity:
+                entity_id = self._entity.get("id")
+                entity_type = self._entity.get("type")
+
             self._task_uid = self._data_retriever.request_thumbnail(
                 value, entity_type, entity_id, self._field_name, load_image=True)
             self._needs_download = False
@@ -297,7 +325,7 @@ class ImageWidget(QtGui.QLabel):
         """
         On failure just display an error and set the toolkit to the error string.
         """
-        if str(uid) == self._task_uid:
+        if uid == self._task_uid:
             self.clear()
             self.setText("Error loading image.")
             self.setToolTip(msg)
@@ -306,9 +334,9 @@ class ImageWidget(QtGui.QLabel):
         """
         Handle the finished download by updating the image the label displays.
         """
-        if str(uid) == self._task_uid:
+        if uid == self._task_uid:
             image = data["image"]
-            self._image_url = data["thumb_path"]
+            self._image_path = data["thumb_path"]
             pixmap = QtGui.QPixmap.fromImage(image)
             self.setPixmap(pixmap)
 
@@ -339,9 +367,9 @@ class ImageWidget(QtGui.QLabel):
         if not self._pixmap:
             return
 
-        if self._image_url:
-            # try to use the path if there is one
-            display_pixmap = QtGui.QPixmap(self._image_url)
+        if self._image_path:
+            # try to use the local path if there is one
+            display_pixmap = QtGui.QPixmap(self._image_path)
         else:
             # may only have the pixmap (delegates)
             display_pixmap = QtGui.QPixmap(self._pixmap)
@@ -350,7 +378,9 @@ class ImageWidget(QtGui.QLabel):
 
         img_dialog = QtGui.QDialog(self)
         img_dialog.setWindowTitle(
-            "Viewing: %s %s" % (self._entity_type, self._field_name)
+            "Viewing: %s %s" % (
+                shotgun_globals.get_type_display_name(self._entity_type),
+                self._field_name)
         )
 
         lbl = QtGui.QLabel()
@@ -380,15 +410,16 @@ class ImageWidget(QtGui.QLabel):
 
     def _update_display(self):
         """
-        If no pixmap, display the button with options.
+        Handles updating the display of the widget when its state changes.
         """
-        if self._pixmap:
-            self.setAlignment(QtCore.Qt.AlignLeft | QtCore.Qt.AlignTop)
-        else:
+        if not self._pixmap:
+            # no pixmap to display
+
             if self._editable:
+                # show the popup button to allow setting a new image
                 self._popup_btn.show()
-                self.setAlignment(QtCore.Qt.AlignLeft | QtCore.Qt.AlignTop)
             else:
+                # not editable, show nothing.
                 self.setText("")
 
     def _upload_image(self):
@@ -403,7 +434,7 @@ class ImageWidget(QtGui.QLabel):
         )[0]
         if file_path:
             self.setPixmap(QtGui.QPixmap(file_path))
-            self._image_url = file_path
+            self._image_path = file_path
             self._value = file_path
             self.value_changed.emit()
 
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -145,8 +145,9 @@ class ImageWidget(QtGui.QLabel):
         self._scaled_width = self.width()
 
         if self._delegate:
-            # in delegate mode, the model will handle fetching the data, including
-            # the thumbnail. So we don't need a data retriever.
+            # in delegate mode. that means this widget is being used to display
+            # multiple entitie's image fields. so setting up a data retriever to
+            # download for a specific entity is pointless.
             self._needs_download = False
         else:
             self._needs_download = True
@@ -166,8 +167,12 @@ class ImageWidget(QtGui.QLabel):
         self._popup_btn = QtGui.QPushButton(self)
         self._popup_btn.setIcon(QtGui.QIcon(":/qtwidgets-shotgun-fields/image_menu.png"))
         self._popup_btn.setFixedSize(QtCore.QSize(18, 12))
-        self._popup_btn.setFocusPolicy(QtCore.Qt.NoFocus)
         self._popup_btn.hide()
+
+        if not self._delegate:
+            # not sure why, but when the widget is being used in a delegate,
+            # this causes editor to close immediately when clicked.
+            self._popup_btn.setFocusPolicy(QtCore.Qt.NoFocus)
 
         # make sure there's never a bg color or border
         self._popup_btn.setStyleSheet("background-color: none; border: none;")
@@ -182,6 +187,9 @@ class ImageWidget(QtGui.QLabel):
         self._view_action = QtGui.QAction("View Image", self)
         self._view_action.triggered.connect(self._show_image)
 
+        self._upload_action = QtGui.QAction("Upload Thumbnail", self)
+        self._upload_action.triggered.connect(self._upload_image)
+
         self._popup_edit_menu = QtGui.QMenu()
         self._popup_edit_menu.addAction(self._clear_action)
         self._popup_edit_menu.addAction(self._replace_action)
@@ -189,6 +197,9 @@ class ImageWidget(QtGui.QLabel):
 
         self._popup_display_menu = QtGui.QMenu()
         self._popup_display_menu.addAction(self._view_action)
+
+        self._popup_upload_menu = QtGui.QMenu()
+        self._popup_upload_menu.addAction(self._upload_action)
 
         self.installEventFilter(self)
 
@@ -216,8 +227,10 @@ class ImageWidget(QtGui.QLabel):
         """
         Clear the widget of the current image, displaying the default value.
         """
+        self._value = None
         self._display_default()
         self.value_changed.emit()
+        self._update_display()
 
     def _display_default(self):
         """
@@ -267,7 +280,10 @@ class ImageWidget(QtGui.QLabel):
         """
 
         if self._editable:
-            menu = self._popup_edit_menu
+            if self._pixmap:
+                menu = self._popup_edit_menu
+            else:
+                menu = self._popup_upload_menu
         else:
             menu = self._popup_display_menu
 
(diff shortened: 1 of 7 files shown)


python/shotgun_fields/image_widget.py
```diff
@@ -36,6 +36,13 @@ def image_url(self):
         """
         return self._image_url
 
+    def clear(self):
```
robblau: docstrings and comments please!  :)

python/shotgun_fields/image_widget.py
```diff
@@ -104,9 +111,12 @@ def resizeEvent(self, event):
         :type event: :class:`~PySide.QtGui.QResizeEvent`
         """
         if self._pixmap:
-            self._scale_pixmap()
-        else:
-            super(ImageWidget, self).resizeEvent(event)
+            scaled_pixmap = self._scale_pixmap(self._pixmap)
```
robblau: Would be good to have a comment about what is going on here.

python/shotgun_fields/image_widget.py
```diff
@@ -116,8 +126,12 @@ def setPixmap(self, pixmap):
         :param pixmap: The pixmap to set as the current pixmap
         :type pixmap: :class:`~PySide.QtGui.QPixmap`
         """
+
         self._pixmap = pixmap
-        self._scale_pixmap()
+        scaled_pixmap = self._scale_pixmap(pixmap)
+        self._scaled_width = scaled_pixmap.width()
+        super(ImageWidget, self).setPixmap(scaled_pixmap)
+        self._update_btn_position()
```
robblau: Same.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -45,36 +46,81 @@ def __init__(self, sg_entity_type, field_name, display_class, editor_class, view
...
             entity_type=self._entity_type,
             field_name=self._field_name,
             bg_task_manager=self._bg_task_manager,
+            delegate=True,
         )
+
+        if self._display_class == self._editor_class:
+            # display and edit classes are the same. we need to make sure
+            # we disable the editing
+            widget.enable_editing(False)
```
robblau: Not sure if this is what the code is doing, but isn't it possible to have the same class for display and editor and want editing?  Like for a checkbox?

python/shotgun_fields/image_widget.py
```diff
@@ -130,13 +144,19 @@ def setup_widget(self):
         self._editable = False
         self._scaled_width = self.width()
 
-        self.setAlignment(QtCore.Qt.AlignLeft | QtCore.Qt.AlignTop)
+        if self._delegate:
+            # in delegate mode. that means this widget is being used to display
+            # multiple entitie's image fields. so setting up a data retriever to
```
manneohrstrom: nit: typo

python/shotgun_fields/image_widget.py
```diff
@@ -163,6 +187,9 @@ def setup_widget(self):
         self._view_action = QtGui.QAction("View Image", self)
         self._view_action.triggered.connect(self._show_image)
 
+        self._upload_action = QtGui.QAction("Upload Thumbnail", self)
```
manneohrstrom: QA note: i found QActions particularily brittle when it comes to GC in some DCCs. We should add to our QA notes for this to carefully test menu operations in our various DCCs.

python/shotgun_fields/image_widget.py
```diff
@@ -215,11 +245,17 @@ def _display_value(self, value):
 
         :param value: The value returned by the Shotgun API to be displayed
         """
-        # queue up the download in the background
-        entity_id = self._entity and self._entity.get("id") or None
-        entity_type = self._entity and self._entity.get("type") or None
-        self._task_uid = self._data_retriever.request_thumbnail(
-            value, entity_type, entity_id, self._field_name, load_image=True)
+
+        if isinstance(value, QtGui.QPixmap):
```
manneohrstrom: this value seems unexpected given the docstring - maybe docstring needs updating?

python/shotgun_fields/image_widget.py
```diff
@@ -215,11 +245,17 @@ def _display_value(self, value):
...
+            self.setPixmap(value)
+
+        if self._needs_download:
+            # queue up the download in the background
+            entity_id = self._entity and self._entity.get("id") or None
```
manneohrstrom: This is hard to read: `X and Y or Z`. I suggest add brackets.

python/shotgun_fields/image_widget.py
```diff
@@ -258,7 +297,7 @@ def _on_worker_failure(self, uid, msg):
         """
         On failure just display an error and set the toolkit to the error string.
         """
-        if uid == self._task_uid:
+        if str(uid) == self._task_uid:
```
manneohrstrom: why is this needed? is the `request_thumbnail` method return value broken?
Please add a comment to explain.

python/shotgun_fields/image_widget.py
```diff
@@ -279,24 +318,53 @@ def _replace_image(self):
         """
         self._upload_image()
 
-    def _scale_pixmap(self):
+    def _scale_pixmap(self, pixmap):
         """
         Scale the pixmap in preparation for display.
         """
-        scaled_pixmap = self._pixmap.scaled(
+        return pixmap.scaled(
```
manneohrstrom: In most other places in the code we use `KeepAspectRatioByExpanding` (see docs [here](https://doc.qt.io/archives/qtjambi-4.5.2_01/com/trolltech/qt/core/Qt.AspectRatioMode.html)). What happens if the image is really tall or wide?

python/shotgun_fields/image_widget.py
```diff
@@ -279,24 +318,53 @@ def _replace_image(self):
...
+
+        Display the image in a transient, modeless QDialog.
         """
-        QtGui.QDesktopServices.openUrl("file:///%s" % (self._image_url))
+
+        # don't continue unless there's somethign to show
+        if not self._pixmap:
+            return
+
+        if self._image_url:
```
manneohrstrom: reading the code, looks like `self._image_url` is actually a path to a local file on disk? Url makes me think it's a `http://...` type url. I think this is confusing - why not say image_path or something? sorry if i misread the code?

One of the reasons to have extra clarity here is because thumb urls are S3 urls with expiry, so they are not valid more than a couple of minutes and should never be serialized or stored, even in memory.

python/shotgun_fields/image_widget.py
```diff
@@ -279,24 +318,53 @@ def _replace_image(self):
...
+            display_pixmap = QtGui.QPixmap(self._image_url)
+        else:
+            # may only have the pixmap (delegates)
+            display_pixmap = QtGui.QPixmap(self._pixmap)
+
+        # construct the dialog, display label, and a button box to close it.
+
+        img_dialog = QtGui.QDialog(self)
+        img_dialog.setWindowTitle(
+            "Viewing: %s %s" % (self._entity_type, self._field_name)
```
manneohrstrom: Will this say `CustomEntity13` instead of `Level`?

python/shotgun_fields/image_widget.py
```diff
@@ -279,24 +318,53 @@ def _replace_image(self):
...
+        )
+
+        lbl = QtGui.QLabel()
```
manneohrstrom: what about aspect ratio scaling? Is this handled by the label? Maybe worth adding some comments.

python/shotgun_fields/image_widget.py
```diff
@@ -312,18 +380,16 @@ def _update_btn_position(self):
 
     def _update_display(self):
         """
-        If no pixmap, display a link to upload if the widget is editable,
-        otherwise display a ``No Image`` string.
+        If no pixmap, display the button with options.
```
manneohrstrom: is part of the docstring missing? This seems like a second line to me, with the summary line missing!

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -45,36 +46,81 @@ def __init__(self, sg_entity_type, field_name, display_class, editor_class, view
...
         """
         widget = self._display_class(
             parent=parent,
             entity_type=self._entity_type,
             field_name=self._field_name,
             bg_task_manager=self._bg_task_manager,
+            delegate=True,
         )
+
+        if self._display_class == self._editor_class:
```
manneohrstrom: probably a stupid question, but what happens if there is inheritance? Is that something we should worry about in this equality check?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -45,36 +46,81 @@ def __init__(self, sg_entity_type, field_name, display_class, editor_class, view
...
+            # display and edit classes are the same. we need to make sure
+            # we disable the editing
+            widget.enable_editing(False)
+
         return widget
 
+    def sizeHint(self, style_options, model_index):
+        """
+        Returns a size hint for the painter widget.
```
manneohrstrom: missing return value and params.

does this mean that the delegates can dynamically compute their size? That's awesome! How does that work?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -45,36 +46,81 @@ def __init__(self, sg_entity_type, field_name, display_class, editor_class, view
...
+        if painter_widget:
+            size_hint = painter_widget.size()
+
+        return size_hint
+
     def _create_editor_widget(self, model_index, style_options, parent):
         """
-        :param model_index:     The index of the item in the model to return a widget for
-        :type model_index:      :class:`~PySide.QtCore.QModelIndex`
+        :param model_index: The index of the item in the model to return a widget for
```
manneohrstrom: missing a description in the docstring.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -92,26 +149,201 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+        :return: ``True``, if the event was handled, ``False`` otherwise.
+        :rtype: ``bool``
+        """
+
+        # Forward mouse clicks to the underlying display widget. This only kicks
+        # in if the editor widget isn't displayed or doesn't process a mouse
+        # event for some reason. If you're having trouble with editors
+        # disappearing, it may be because they can't receive focus or aren't
+        # handling a mouse click.
+        if event.type() == QtCore.QEvent.MouseButtonRelease:
```
manneohrstrom: sorry for being stupid but i really don't understand what case we are trying to handle here. We are tracking that the mouse button is released inside the widget? Maybe worth explaining a bit more when this can happen? Is it when you drag the mouse from the outside into the widget or something like that?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -92,26 +149,201 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+            self._forward_mouse_event(event, index)
+            return True
+
+        return False
+
+    def _forward_mouse_event(self, mouse_event, index):
+        """
+        Forward the mouse event to the display widget to simulate
+        interacting with the widget. This is necessary since the delegate only
+        paints the widget in the view rather than being and actual widget
```
manneohrstrom: nit: and -> an

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -92,26 +149,201 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+    if not src_index or not src_index.isValid():
+        # invalide index, do nothing
+        return
+
+    # special case for image fields
+    if widget._field_name == "image":
+        primary_item = src_index.model().item(src_index.row(), 0)
+        icon = primary_item.icon()
+        if icon:
+            widget.set_value(icon.pixmap(QtCore.QSize(256, 256)))
```
manneohrstrom: should this be a constant inside the image handling widget?

python/views/widget_delegate.py
```diff
@@ -85,6 +85,10 @@ def _get_painter_widget(self, model_index, parent):
         :returns:           A QWidget to be used for painting the current index
         :rtype:             :class:`~PySide.QtGui.QWidget`
         """
+
+        if not model_index.isValid():
+            return None
```
manneohrstrom: a comment here would be helpful. Why are we exiting early?

python/views/widget_delegate.py
```diff
@@ -114,6 +118,10 @@ def _create_editor_widget(self, model_index, style_options, parent):
         """
         # the default implementation just calls _create_widget for backwards
         # compatibility.
+
+        if not model_index.isValid():
+            return None
```
manneohrstrom: a comment here would be helpful. Why are we exiting early? What's the use case when the index isn't valid?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -92,26 +149,201 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+
+    :param widget: The widget to set the value for
+    :type parent: :class:`~PySide.QtGui.QWidget`
+    :param model_index: The index of the model where the data comes from
+    :type model_index: :class:`~PySide.QtCore.QModelIndex`
+    """
+
+    src_index = _map_to_source(model_index)
+    if not src_index or not src_index.isValid():
+        # invalide index, do nothing
```
manneohrstrom: nit: typo

why are indices invalid?

manneohrstrom:
I did a quick CR pass - hope it's helpful.

One general comment I have is that there is very little debug logging. Please consider the use case that something is not working for a client and we need to support them - in that case it's helpful to ask them to turn on debug and at least get some sort of scaffoldy set of debug logging info back to help pinpoint what is happening, what the data looks like etc.

python/shotgun_fields/image_widget.py
josh-t: added.

python/shotgun_fields/image_widget.py
josh-t: done.

python/shotgun_fields/image_widget.py
josh-t: done.

python/shotgun_fields/image_widget.py
josh-t: fixed.

python/shotgun_fields/image_widget.py
josh-t: will do!

python/shotgun_fields/image_widget.py
josh-t: good catch. updated docstring.

python/shotgun_fields/image_widget.py
josh-t: fixed both of these: `entity_id` and `entity_type`

python/shotgun_fields/image_widget.py
josh-t: I don't remember why I added this... I just tested, and things seem to work without. I'm guessing I was testing/debugging something on a hunch that proved incorrect... I've removed this from both methods.

python/shotgun_fields/image_widget.py
josh-t: I believe `KeepAspectRatioByExpanding` will cause part of the image to be clipped in tall/wide scenarios. My expectation is that user's would prefer to see the whole image, but I'm happy to switch if it should clip. :)

python/shotgun_fields/image_widget.py
josh-t: Yes, thanks! I have deprecated the `image_url` property and added `image_path` instead. I got rid of the previous call to `QDesktopServices` because of the expiry issue. Now it pops up a Qt dialog displaying the downloaded image path if there is one. Hope that makes sense!

python/shotgun_fields/image_widget.py
josh-t: Fixed with call to `shotgun_globals.get_type_display_name()`! thanks!

python/shotgun_fields/image_widget.py
josh-t: The popup dialog is just the label with an `Ok` button. My assumption was that this would always be a relatively small image, but maybe that's wrong. Does SG limit/scale the `image` field uploads on the backend? What do you think is the right way to handle this? I'm not tied to this at all. :)

python/shotgun_fields/image_widget.py
josh-t: This is strictly a poorly worded summary of what the method does... fixed and cleaned up the method too.

python/shotgun_fields/shotgun_field_delegate.py
josh-t: you mean if the editor is a subclass of display or vice versa? in that case, those are still 2 separate widgets that will be instantiated for display/editing so i think this is ok.

python/shotgun_fields/shotgun_field_delegate.py
josh-t: Yes, having the same class for both is possible and encouraged. We only really want the display class here but this call ensures that editing is disabled so that the delegate isn't drawing the widget in its edit state. That make sense? I updated the comment too to make that clearer. :)

python/shotgun_fields/shotgun_field_delegate.py
josh-t: Updated the docstring...

This just returns the size hint for the widget to be painted. So it doesn't really know anything about the view, only a reasonable size for the widget itself (if provided). Currently the `model_index` isn't used to populate the delegate with any data, but presumably we could change it to do so. That could be used by the widgets themselves to drive their `sizeHint`. Thoughts?

python/shotgun_fields/shotgun_field_delegate.py
josh-t: fixed.

python/shotgun_fields/shotgun_field_delegate.py
josh-t: The primary use for this is labels displaying clickable links (entity, multi-entity, etc). By default, they're painted into the view via the delegate, so you can't interact with them. There were some suggestions online how to work around this that seemed really complex. This is a solution @robblau suggested which I tried and it seems to work... and is much simpler! Basically, detect a mouse click (release is all we have really) in the delegate, populate the underlying widget with the data from the index, then forward the event to the widget. The result is a simulation of clicking on the actual widget. That make sense?

python/shotgun_fields/shotgun_field_delegate.py
josh-t: not sure. overly cautious?

python/shotgun_fields/shotgun_field_delegate.py
josh-t: Yes, probably. Let me figure out what SG does as far as limiting thumbnails and I'll update.

python/views/widget_delegate.py
josh-t: added.

python/views/widget_delegate.py
josh-t: I tend to always include this check in methods that accept model index args since who knows what kind of craziness can go on in a Qt model, proxy, etc. I'll add a comment here too.

python/shotgun_fields/image_widget.py
manneohrstrom: maybe do a quick test with a collection of different aspect ratio thumbs and see what it looks like when displayed side by side!

python/shotgun_fields/image_widget.py
manneohrstrom: Yep, the thumbnail is indeed scaled at the backend so probably safe! Maybe worth quickly checking with some really high and/or tall images just to be safe!

python/shotgun_fields/shotgun_field_delegate.py
manneohrstrom: Ah, ok. So basically same as before :) Sounds good to start with. I think at some point in the future would be awesome to be able to compute say the size of a given rendered text and then base the height on that sort of measurement. But for now, this is definitely sufficient - that sort of work should be costed and ticketed separately i think! (maybe you can make a ticket?)

python/shotgun_fields/shotgun_field_delegate.py
manneohrstrom: sounds cool! I think if you merge that reply into a code comment that would be great!

python/shotgun_fields/image_widget.py
josh-t: Agreed. Here's a test: 

`KeepAspectRatio`
----
widget:
screen shot 2016-07-19 at 8 11 34 am
delegate (tall & wide):
screen shot 2016-07-19 at 8 09 20 am

`KeepAspectRatioByExpanding`
----
widget:
screen shot 2016-07-19 at 8 11 13 am
delegate (tall & wide):
screen shot 2016-07-19 at 8 11 07 am

Obviously the way it's currently implemented, using the stand alone display/edit widget looks better... though I'm sure it could be smarter. As for the delegates, I think wide images are better when displaying the full image. The tall image is equally useless in either case I think. :)

I'm inclined to keep it as-is unless you feel strongly it should be changed.

python/shotgun_fields/shotgun_field_delegate.py
josh-t: Agreed. I think all it would require is to actually set the data on the paint widget here and then ensure the widgets themselves return proper `sizeHints` based on their contents. I've created a ticket internally (#37522).

python/shotgun_fields/shotgun_field_delegate.py
josh-t: haha! done!

python/shotgun_fields/image_widget.py
manneohrstrom: yes, totally agree let's keep as is! Sorry for the noise, thanks for the test!

python/shotgun_fields/image_widget.py
manneohrstrom: nice!

python/shotgun_fields/shotgun_field_delegate.py
manneohrstrom: brilliant! this is so exciting.

python/views/widget_delegate.py
manneohrstrom: that sounds great - personally, i think it's good to be overly cautios - that's how i tend to code too - recently, i found myself going back to code i had previously written thinking "did i put this in just because i am overly cautious or is it addressing a specific use case?". For that sort of situation, it's really great to have a quick comment to help explain.

PR #28, opened 2016-07-01
#37270: Adds the ability to disable the clickable "hand" cursor on user thumbs.

thebeeland:
This is just a new configuration option for the activity stream widget. In the Shotgun Panel this is used to navigate to the User entity associated with the thumbnail, but in other applications of this widget we want to be able to disable that capability.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -72,6 +72,7 @@ class ActivityStreamWidget(QtGui.QWidget):
         self._allow_screenshots = True
         self._show_sg_stream_button = True
         self._version_items_playable = True
+        self._clickable_user_icons = True
         
         # apply styling
         self._load_stylesheet()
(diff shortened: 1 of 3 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -158,6 +159,30 @@ def note_widget(self):
         """
         return self.ui.note_widget
 
+    def _get_clickable_user_icons(self):
+        """
+        Whether user icons in the activity stream display as clickable.
```
manneohrstrom: Can we also mention the default value (e.g. hand)?
What happens when you get the arrow? Are mouse clicks ignored?

python/activity_stream/activity_stream.py
```diff
@@ -158,6 +159,30 @@ def note_widget(self):
...
+        hovered over the icons, otherwise the default arrow cursor will be
+        used.
+        """
+        return self._clickable_user_icons
+
+    def _set_clickable_user_icons(self, state):
+        self._clickable_user_icons = bool(state)
+
+        for widget in self._activity_stream_data_widgets.values():
+            if isinstance(widget, NoteWidget):
```
manneohrstrom: what does it mean that it only applies to NoteWidgets? Should this be reflected in the docs?

python/activity_stream/widget_note.py
```diff
@@ -54,10 +54,32 @@ def __init__(self, parent):
...
-        self.ui.user_thumb.entity_requested.connect(lambda entity_type, entity_id: self.entity_requested.emit(entity_type, entity_id))    
+        self.ui.user_thumb.entity_requested.connect(lambda entity_type, entity_id: self.entity_requested.emit(entity_type, entity_id))
+
+    ##############################################################################
+    # properties
+
+    @property
+    def user_thumb(self):
+        """
+        The user thumbnail widget.
```
manneohrstrom: what's the type returned here? Is that object currently part of the public API? Is it documented?

python/activity_stream/widget_note.py
```diff
@@ -54,10 +54,32 @@ def __init__(self, parent):
...
+        """
+        return self.ui.user_thumb
 
     ##############################################################################
     # public interface
+
+    def set_user_thumb_cursor(self, cursor):
+        """
+        Sets the cursor displayed when hovering over the user
```
manneohrstrom: This seems like an odd addition to me, it's not the kind of public interface member I would expect on a Shotgun note widget :) What's the use case here? Sorry, just curious. At the very least, it would be useful to have an example in the docstring/sphinx docs to explain when this method is useful.

PR #30, opened 2016-07-25
#36274: Extracts Shotgun Review's version details widget into the framework.

thebeeland:
This is a lot of code, much of which has been previously reviewed as part of the Shotgun Review app work. Below are screens of both the version details app, as well as the shotgun panel as it looks using this framework code in Maya 2016.

This also addresses Shotgun Panel UI regressions outlined in #37385.

capture

capture2


python/activity_stream/activity_stream.py
```diff
@@ -150,6 +158,29 @@ def destroy(self):
...
     @property
+    def note_threads(self):
+        """
+        The currently loaded note threads, keyed by Note entity id and
+        containing a list of Shotgun entity dictionaries.
+
+        Example structure containing a single Note entity:
+            6038: [
+                {
+                    'content': 'This is a test note.',
```
manneohrstrom: is the first item always a Note entity and then there are reply entities following? Could there be attachments in this list?

python/activity_stream/activity_stream.py
```diff
@@ -150,6 +158,29 @@ def destroy(self):
     # properties
 
     @property
+    def note_threads(self):
+        """
+        The currently loaded note threads, keyed by Note entity id and
```
manneohrstrom: nitpick: could be worth clarifying what "currently loaded" means here.

python/activity_stream/activity_stream.py
```diff
@@ -255,16 +286,69 @@ def _set_version_items_playable(self, state):
...
+        _get_notes_are_selectable,
+        _set_notes_are_selectable,
+    )
         
     ############################################################################
     # public interface
+
+    def get_note_attachments(self, note_id):
+        """
+        Gets the Attachment entities associated with the given Note
```
manneohrstrom: are these also just the "currently loaded" ones?

python/activity_stream/activity_stream.py
```diff
@@ -412,6 +496,39 @@ def load_data(self, sg_entity_dict):
...
+        :type modal:    bool
+        """
+        note_dialog = NoteInputDialog(parent=self)
+        note_dialog.entity_created.connect(self._on_entity_created)
+        note_dialog.data_updated.connect(self.rescan)
+        note_dialog.set_bg_task_manager(self._task_manager)
+        note_dialog.set_current_entity(self._entity_type, self._entity_id)
+
+        if modal:
+            note_dialog.exec_()
```
manneohrstrom: if the return value of any interest here?

python/activity_stream/activity_stream.py
```diff
@@ -412,6 +496,39 @@ def load_data(self, sg_entity_dict):
...
+        else:
+            note_dialog.show()
+
+
+    def rescan(self, force_activity_stream_update=False):
+        """
+        Triggers a rescan of the current activity stream data.
+
+        :param force_activity_stream_update:    If True, will force a requery
```
manneohrstrom: sounds useful! is it hooked up to a refresh button or something like that?

python/activity_stream/activity_stream.py
```diff
@@ -692,6 +824,18 @@ def _process_new_note(self, activity_id, note_id):
...
 
+            # Ordering here is important. An example of why is that higher
+            # level widgets/apps might be registering a newly-arrived note
+            # with some host application and also listening for selection
+            # signals to notify the same host application of that having
+            # occurred. If the host application doesn't know the note exists,
+            # then telling it the note is selected won't do any good.
+            self.note_arrived.emit(note_id)
+            widget.set_selected(True)
+            return
```
manneohrstrom: is it clearer to have an else instead?

python/activity_stream/data_manager.py
```diff
@@ -109,9 +110,35 @@ def __init__(self, parent):
...
+        # a forced requery of activity stream data during rescan.
+        self._force_activity_stream_update = False
                 
         # set up defaults
         self.__reset()
+
+    @property
+    def note_threads(self):
+        """
+        The list of currently-loaded note threads, keyed by Note entity id.
```
manneohrstrom: same comments as for the activity widget note threads property :)

python/activity_stream/data_manager.py
```diff
@@ -217,7 +245,7 @@ def load_activity_data(self, entity_type, entity_id, limit=200):
         sorted_keys = sorted(self._activity_data.keys())
         return sorted_keys
 
-    def rescan(self):
+    def rescan(self, force_activity_stream_update=False):
```
manneohrstrom: docstring param!

python/activity_stream/data_manager.py
```diff
@@ -560,34 +589,47 @@ def __db_insert_activity_updates(self, connection, cursor, entity_type, entity_i
...
+
                 # first insert event
-                sql = """
-                    INSERT INTO activity(activity_id, payload, created_at) 
-                    SELECT ?, ?, datetime('now')
-                    WHERE NOT EXISTS(SELECT activity_id FROM activity WHERE activity_id = ?);                
-                 """
+                if self._force_activity_stream_update:
+                    sql = """
+                        INSERT INTO activity(activity_id, payload, created_at)
```
manneohrstrom: does this mean we are double inserting records? Should there be some stage where things are deleted from the db too?

python/activity_stream/resources/build_resources.sh
```diff
@@ -36,16 +36,16 @@ function build_res {
 
 # build UI's:
 echo "building user interfaces..."
-# build_ui activity_stream_widget
+build_ui activity_stream_widget
```
manneohrstrom: :)

python/activity_stream/widget_note.py
```diff
@@ -32,6 +32,8 @@ class NoteWidget(ActivityStreamBaseWidget):
     Widget that represents a Note. This widget in turn contains
     replies and attachments.
     """
+
+    selection_changed = QtCore.Signal(bool, int)
```
manneohrstrom: please explain the params in a comment

python/activity_stream/widget_note.py
```diff
@@ -49,23 +51,87 @@ def __init__(self, parent):
         self._general_widgets = []
         self._reply_widgets = []
         self._attachment_group_widgets = {}
-                
+        self._selected = False
+        self._attachments = []
+        self._show_note_links = True
+
+        self.setStyleSheet("#frame { border: 1px solid transparent }")
```
manneohrstrom: is this so that it doesn't "jump" when it gets highlighted? May be worth adding a comment explaining.

python/activity_stream/widget_note.py
```diff
@@ -49,23 +51,87 @@ def __init__(self, parent):
...
 
     @property
+    def attachments(self):
+        """
+        Returns a list of attachment entities as returned from the
+        Shotgun Python API.
+
+        Example:
+        [
+            {'attachment_links': [{'id': 6043,
```
manneohrstrom: these examples are super helpful!

python/activity_stream/widget_note.py
```diff
@@ -175,15 +247,43 @@ def add_replies(self, replies_and_attachments):
...
+        :param state:   Whether the widget is to be selected or deselected.
+        :type state:    bool
+        """
+        if bool(state) == self.selected:
+            return
+
+        self._selected = bool(state)
+
+        if self.selected:
+            color = self.palette().color(
```
manneohrstrom: nice!

python/models/shotgun_sort_filter_proxy_model.py
```diff
@@ -0,0 +1,233 @@
...
+    "tk-framework-shotgunutils",
+    "shotgun_model",
+)
+
+shotgun_globals = sgtk.platform.import_framework(
+    "tk-framework-shotgunutils",
+    "shotgun_globals",
+)
+
+class ShotgunSortFilterProxyModel(QtGui.QSortFilterProxyModel):
```
manneohrstrom: please make sure this gets added to the sphinx docs

python/models/shotgun_sort_filter_proxy_model.py
```diff
@@ -0,0 +1,233 @@
...
+    """
+    A sort/filter proxy model that handles sorting and filtering
+    data in a ShotgunModel by given Shotgun fields on the entities
+    stored therein.
+
+    :ivar filter_by_fields:         A list of string Shotgun field names to filter on.
```
manneohrstrom: This feels a little unusual - normally in toolkit we use properties to expose interfaces. Any reason not keeping it consistent?

python/models/shotgun_sort_filter_proxy_model.py
```diff
@@ -0,0 +1,233 @@
...
+                    sort_by_field,
+                    sortable=True,
+                )
+                right_data = self._get_processable_field_data(
+                    sg_right,
+                    sort_by_field,
+                    sortable=True,
+                )
+            except KeyError:
+                continue
```
manneohrstrom: a comment here why this is ok would be helpful

python/models/simple_tooltip_model.py
```diff
@@ -0,0 +1,35 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+
+shotgun_model = sgtk.platform.import_framework(
+    "tk-framework-shotgunutils",
+    "shotgun_model",
+)
+
+class SimpleTooltipModel(shotgun_model.SimpleShotgunModel):
```
manneohrstrom: - Some info explaining how this is being used and when it is useful would be great
- please add to sphix
- Should this be in shotgun utils, alongside the simple shotgun model?

python/models/simple_tooltip_model.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Sets the tooltip of the given item.
+
+        :param item:    Shotgun model item that requires a tooltip.
+        :param sg_item: Dictionary of the entity associated with the Shotgun
+                        model item.
+        """
+        item.setToolTip(
+            "%s: %s" % (
+                sg_item.get("type", "Entity Type Unknown"),
```
manneohrstrom: If the name isn't set in Shotgun, the sg_item will probably look like this `{ ...., "code": None }`.
In this case the code will not work. 

The pattern i normally use is 

```
sg_item.get("code") or "Entity Name Unknown"
```

just to cover all cases.

The output feels very technical to me - examples:

- Shot: Foo Bar
- Entity Type Unknown: Foo Bar
- Shot: None
- Project: Entity Name Unknown
- CutomEntity13: Foo Bar
- Entity Type Unknown: Entity Name Unknown

I suggest:

- don't assume that `code` is the name field - use sg globals to determine the appropriate name field. Projects and tasks for example do not use `code`.
- the logic doesn't handle custom entities - again, use sg globals to resolve
- the output itself feels like it could be polished.

I suggest the following output (same cases as above)

- Shot: Foo Bar --> Shot Foo Bar
- Entity Type Unknown: Foo Bar --> This is an anomaly and no tooltip should be set - log a warning.
- Shot: None --> Unnamed Shot
- Project: Entity Name Unknown  --> Unnamed Project
- CutomEntity13: Foo Bar --> Level Foo Bar (Display name for CustomEntity13 is Level)
- Entity Type Unknown: Entity Name Unknown --> This is an anomaly and no tooltip should be set - log a warning.

python/version_details/resources/resources.qrc
```diff
@@ -0,0 +1,26 @@
...
+    <file>right_arrow_large.png</file>
+    <file>right_arrow.png</file>
+    <file>right_arrow_white.png</file>
+    <file>panel_empty_background.png</file>
+    <file>dock.png</file>
+    <file>dock_hover.png</file>
+    <file>sort_down_on.png</file>
+    <file>sort_up_on.png</file>
+    <file>sort_down.png</file>
+    <file>sort_up.png</file>
```
manneohrstrom: make sure to include source images for all icons etc. We normally push a graffle file, but anything works. :)

python/version_details/version_details.py
```diff
@@ -0,0 +1,1213 @@
...
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import os
```
manneohrstrom: please add some docs to sphinx.

python/version_details/version_details.py
```diff
@@ -0,0 +1,1213 @@
...
+    def setParent(self, parent):
+        """
+        Calls the base class' method of the same name, and then checks
+        to see if the parent is a dock widget. If it is, then a custom
+        title bar is made visible.
+        """
+        super(VersionDetailsWidget, self).setParent(parent)
+
+        try:
+            self.parent().dockLocationChanged.connect(self._dock_location_changed)
```
manneohrstrom: will this work in maya too?

python/shotgun_menus/entity_field_menu.py
```diff
@@ -78,6 +78,10 @@ def __init__(self, sg_entity_type, parent=None, bg_task_manager=None):
 
         :param bg_task_manager: The task manager the menu will use if it needs to run a task
         :type bg_task_manager: :class:`~task_manager.BackgroundTaskManager`
+
+        :param int project_id: The project entity id to use when querying fields lists. If
+                               None, the current context's project will be used if one is
+                               set, otherwise the "site" level fields will be queried.
```
manneohrstrom: i thin this is similar to this comment:

https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/46#discussion_r71131214

If project id is None, aren't we using the site wide settings?

python/shotgun_menus/entity_field_menu.py
```diff
@@ -221,7 +234,12 @@ def entity_filter(et):
                     bubble_fields[field] = {
                         "name": display_name,
                         "valid_types": entity_types,
-                        "valid_type_names": [shotgun_globals.get_type_display_name(et) for et in entity_types],
+                        "valid_type_names": [
+                            shotgun_globals.get_type_display_name(
+                                et,
+                                project_id=self._project_id,
+                            ) for et in entity_types
+                        ],
```
manneohrstrom: the `except Exception` below feels a bit brutal? would it make sense to reduce the scope, either in terms of lines of code or in terms of faults to trap? Not saying we have to, just pointing it out :)

python/shotgun_fields/widget_metaclass.py
```diff
@@ -0,0 +1,175 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk.platform.qt import QtGui
+from .shotgun_field_manager import ShotgunFieldManager
+
+
+# use type to pull the metaclass for QWidget since Shiboken is not always directly available
+# and it abstracts out any difference between PySide and PyQt
+class ShotgunFieldMeta(type(QtGui.QWidget)):
```
manneohrstrom: sphinx?

python/shotgun_entities/card_widget.py
```diff
@@ -0,0 +1,464 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from collections import OrderedDict
+
+import sgtk
+from sgtk.platform.qt import QtCore, QtGui
+from .ui.card_widget import Ui_ShotgunEntityCardWidget
+
+class ShotgunEntityCardWidget(QtGui.QWidget):
```
manneohrstrom: sphinx!

python/shotgun_entities/card_delegate.py
```diff
@@ -0,0 +1,249 @@
...
+from sgtk.platform.qt import QtCore, QtGui
+from .card_widget import ShotgunEntityCardWidget
+
+views = sgtk.platform.current_bundle().import_module("views")
+shotgun_model = sgtk.platform.import_framework(
+    "tk-framework-shotgunutils",
+    "shotgun_model",
+)
+
+class ShotgunEntityCardDelegate(views.EditSelectedWidgetDelegate):
```
manneohrstrom: sphinx?

python/shotgun_entities/card_delegate.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    A delegate wrapper for :class:`~ShotgunEntityCardWidget`.
+
+    :ivar show_labels:              Whether to show labels for any Shotgun
```
manneohrstrom: so what's the story here? maybe we can discuss a bit? We used to have widgets as part of the qtwidgets fw, but then we pulled them out because we felt it was premature.

manneohrstrom:
Added comments!

manneohrstrom:
Looking very exciting! :)

python/activity_stream/activity_stream.py
thebeeland: Probably, yes. This specific feature is pretty rough right now and is up for a polish pass before Dynamite's Shotgun integration is released.

python/activity_stream/activity_stream.py
thebeeland: The way I'm using it is to force a recache of the stream data after there's been an update to the Shotgun data after I know the widget has already cached. In my case it's adding new file attachments to a Note entity, but there are other uses. If/When we expose the ability to edit fields on an entity (like a Note) from this widget (or a higher-level widget like the version details widget) we'll need a way to invalidate the sql lite cache for that entity.

python/activity_stream/widget_note.py
thebeeland: Yeah, that's the reason for the transparent border.

python/models/shotgun_sort_filter_proxy_model.py
thebeeland: No specific reason other than brevity. I'm fine with it being a property instead for the sake of consistency.

python/version_details/version_details.py
thebeeland: Likely not. It's a bit of a hack to get the right look in RV without breaking when used in other host applications.

thebeeland: The net result is just that it gets the default dock title bar styling when used in a DCC where this hack doesn't work.

python/shotgun_entities/card_delegate.py
thebeeland: Not sure I understand what you're saying. We can discuss.

python/activity_stream/activity_stream.py
manneohrstrom: Ah, so we are in a holding pattern now, and won't release these changes until the dynamite integration is released? So that there can be tweaks and bug fixes? And then, the next step is to upgrade RV to use these official widgets instead? That sounds cool!

python/activity_stream/data_manager.py
```diff
@@ -247,6 +275,7 @@ def rescan(self):
                     "entity_id": self._entity_id,
                     "highest_id": highest_id
                     }
+            self._force_activity_stream_update = force_activity_stream_update
```
manneohrstrom: Why not pass this as part of `data`?

python/version_details/version_details.py
manneohrstrom: at the very least we need some comments explaining this. If there was a way of keeping the framework DCC agnostic that would be amazing too.

python/activity_stream/activity_stream.py
thebeeland: As long as we're happy with the programmer interface I think we can go ahead and get this released. As such, I'll get the return value sorted out now. ;)

thebeeland: After another, look, I don't think we should worry about it here. The activity stream widget emits an entity_created signal, which is what callers should be listening for rather than relying on the return value of the dialog. I'll be noting that in the docs.

python/version_details/resources/resources.qrc
thebeeland: In this case, all of these are simple graphical shapes with no real source image to speak of. Just some very basic Photoshop work to paint a single layer containing the shape.

python/version_details/version_details.py
thebeeland: I'll just remove this for now. RV currently isn't using this code, so nothing is taking advantage of it. I'll address the RV-specific problem in that app once it is using this widget.

python/shotgun_menus/entity_field_menu.py
thebeeland: It does seem brutal, and that's a really big chunk of logic to wrap in such an all-encompassing try/except. Unfortunately, I'm not familiar enough with that chunk of logic to feel comfortable changing it (It's Rob's/Josh's stuff).

python/shotgun_entities/card_delegate.py
thebeeland: It's been pushed under version_details, as discussed.

PR #31, opened 2016-08-05
#36274: Adds attachments filtering mechanism and new note autoselection.

thebeeland:
Feature polish for SG3DR. Allows Note attachments to be filtered out of the attachment group widget based on a regex match against the attachment name. Also autoselects newly-created Notes if note selection is activated in the activity stream widget.

Bug fix: forces an sqlite cache update from SG when new attachments to a Note entity are made.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -86,6 +86,7 @@ class ActivityStreamWidget(QtGui.QWidget):
         self._show_note_links = True
         self._highlight_new_arrivals = True
         self._notes_are_selectable = False
+        self._attachments_filter = None
         
         # apply styling
         self._load_stylesheet()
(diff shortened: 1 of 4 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -372,6 +374,18 @@ def _set_notes_are_selectable(self, state):
...
     )
+
+    def _get_attachments_filter(self):
+        """
+        If set to a compiled regular expression, attachment file names that match
+        will be filtered OUT and NOT shown.
+        """
+        return self._attachments_filter
+
+    def _set_attachments_filter(self, regex):
```
robblau: Docstring would be good.  What is regex?  What is it going to be used for?

thebeeland: For properties, sphinx pulls docstrings from the getter and ignore the setter. I'll add a comment to the setter for people reading the code, and will expand the getter docstring a bit.

PR #32, opened 2016-08-08
#37714 Improved activity stream note creation in site context

manneohrstrom:
This adds logic to the activity stream widget and note input widget to better handle the state when no project is available in the context (e.g. a "site" context).

Notes that are created require a project to be set and there are some cases when a project isn't available. Normally, we pull the project from the context but in the site context situation this is not an option. Instead, the note creation widget now tries to look for a project on the entity itself and if that fails in errors. The activity stream in turn hides the note input widgets in these cases.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -427,9 +427,11 @@ class ActivityStreamWidget(QtGui.QWidget):
 
         # if the project context is None, the entity is a non project entity and it's NOT
         # the project entity itself, we don't have access to any project state.
-        no_project_available = self._bundle.context.project is None \
-                               and is_non_project_entity_type \
-                               and self._entity_type != "Project"
+        no_project_available = (
+            self._bundle.context.project is None and
+            is_non_project_entity_type and
+            self._entity_type != "Project"
+        )
 
         # also disable note creation in the case we have a site context
         # and a non-project entity
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -23,7 +23,9 @@ from .dialog_reply import ReplyDialog
 from .data_manager import ActivityStreamDataHandler
 from .overlaywidget import SmallOverlayWidget
 from ..note_input_widget import NoteInputDialog
- 
+
+shotgun_globals = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_globals")
+
 class ActivityStreamWidget(QtGui.QWidget):
     """
     QT Widget that displays the Shotgun activity stream for an entity.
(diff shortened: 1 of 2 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -418,8 +420,25 @@ def load_data(self, sg_entity_dict):
...
         # inconsistent for users, we need to disable to note dialog for 
         # these cases
+
+        # note - this may return [] if shotgun globals aren't yet cached
+        schema_fields = shotgun_globals.get_entity_fields(self._entity_type)
+        is_non_project_entity_type = len(schema_fields) > 0 and "project" not in schema_fields
+
+        # if the project context is None, the entity is a non project entity and it's NOT
+        # the project entity itself, we don't have access to any project state.
+        no_project_available = self._bundle.context.project is None \
```
thebeeland: [nitpicky] I believe you can wrap these in () and not have to \ the newlines.

python/note_input_widget/widget.py
```diff
@@ -456,23 +472,38 @@ def _async_submit_note(self, sg, data):
...
             note_names = [x["name"] for x in note_links]
             title += " on %s" % (", ".join(note_names))
 
+        # step 3 - handle project gracefully
+        if project is None:
+            # attempt to pull it from the entity link
+            if entity_project_link is None:
+                # there is no associated project - likely this is a note
+                # on a non-project entity created in the site ctx
+                raise TankError(
```
thebeeland: I'm sure this is handled gracefully in higher-level code, right? Do we pop up a dialog or something, or does it just go to the console?

python/note_input_widget/widget.py
```diff
@@ -341,79 +341,88 @@ def _async_submit_note(self, sg, data):
...
-            # if we are adding a note to a version, link it with the version 
+            # if we are adding a note to a version, link it with the version
             # and the entity that the version is linked to.
-            # if the version has a task, link the task to the note too.  
-            sg_version = sg.find_one("Version", 
-                                     [["id", "is", entity_link["id"] ]], 
+            # if the version has a task, link the task to the note too.
+            sg_version = sg.find_one("Version",
+                [["id", "is", entity_link["id"] ]],
                                      ["entity", "sg_task", "cached_display_name", "project"])
```
thebeeland: Indentation looks weird now (at least in the github diff it does).

python/note_input_widget/widget.py
```diff
@@ -341,79 +341,88 @@ def _async_submit_note(self, sg, data):
...
-            
+
         elif entity_link["type"] == "Task":
             # if we are adding a note to a task, link the note to the entity that is linked to the
             # task. The link the task to the note via the task link.
-            sg_task = sg.find_one("Task", 
-                                  [["id", "is", entity_link["id"] ]], 
+            sg_task = sg.find_one("Task",
+                [["id", "is", entity_link["id"] ]],
                                   ["entity", "project"])
```
thebeeland: Weird indentation.

python/note_input_widget/widget.py
manneohrstrom: This is a tricky one. Because it runs in a worker thread, it makes the reporting more complicated. The raised error will bubble up to the user - popping up a UI is complicated and I also looked at trying to determine it before submitting the work to the bg thread, but in order to do that you would have to query shotgun - but all sg queries happening in the bg - so chicken and egg! This (in conjunction with the widget calling this handling it correctly) seemed like a reasonable compromise.

python/activity_stream/activity_stream.py
manneohrstrom: i kind of prefer this syntax too - switched.

python/note_input_widget/widget.py
manneohrstrom: weird! not sure what happened there... fixed!

python/note_input_widget/widget.py
manneohrstrom: i think this may have happened as i was merging in some changes from master... fixed now.

PR #34, opened 2016-08-12
Replaced OrderedDict with a pure Python implementation of the same.

thebeeland:
OrderedDict is requires Python 2.7 and therefore shouldn't be used in our frameworks.

### DIFF
--- a/python/version_details/shotgun_entities/card_widget.py
+++ b/python/version_details/shotgun_entities/card_widget.py
@@ -465,6 +465,11 @@ class _OrderedDict(object):
     def __init__(self, **kwargs):
         """
         Constructor. Emulates the behavior of the dict() type.
+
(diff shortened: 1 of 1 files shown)


python/version_details/shotgun_entities/card_widget.py
```diff
@@ -462,3 +457,69 @@ def field_widgets(self):
...
 
+class _OrderedDict(object):
+    """
+    An OrderedDict-like class. This is implemented here in order to maintain
+    backwards compatibility with pre-2.7 releases of Python.
+    """
+    def __init__(self, **kwargs):
+        """
+        Constructor. Emulates the behavior of the dict() type.
+        """
```
josh-t: might be worth pointing out here that the order of the supplied dict will not be maintained.

PR #35, opened 2016-08-16
#37784: Fields widgets are now editable.

thebeeland:
This makes use of editable shotgun_fields widgets to allow a user to modify certain fields on a Version that is active in the version_details widget.

### DIFF
--- a/python/shotgun_fields/shotgun_field_editable.py
+++ b/python/shotgun_fields/shotgun_field_editable.py
@@ -90,6 +90,33 @@ class ShotgunFieldEditable(QtGui.QStackedWidget):
         # make sure display widget follows
         self._display.enable_editing(self._enabled)
 
+    def get_entity(self):
+        """
+        Returns the entity of the widget (display or editor) currently being shown.
+        """
+        if self.currentWidget() == self._display:
+            return self._display.display_widget.get_entity()
+        else:
+            return self._editor.edit_widget.get_entity()
+
+    def get_entity_type(self):
+        """
+        Returns the entity type of the widget (display or editor) currently being shown.
+        """
+        if self.currentWidget() == self._display:
+            return self._display.display_widget.get_entity_type()
+        else:
+            return self._editor.edit_widget.get_entity_type()
+
+    def get_field_name(self):
+        """
+        Returns the field name of the widget (display or editor) currently being shown.
+        """
+        if self.currentWidget() == self._display:
+            return self._display.display_widget.get_field_name()
+        else:
+            return self._editor.edit_widget.get_field_name()
+
     def get_value(self):
         """
         Returns the value of the widget (display or editor) currently being shown.
--- a/python/shotgun_fields/shotgun_field_editable.py
+++ b/python/shotgun_fields/shotgun_field_editable.py
@@ -12,16 +12,18 @@ import sgtk
 from sgtk.platform.qt import QtCore, QtGui
 from .ui import resources_rc
 
-# TODO: states for "updating in SG" and "failed to update in SG"
 
 class ShotgunFieldEditable(QtGui.QStackedWidget):
     """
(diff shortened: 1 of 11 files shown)


python/shotgun_menus/entity_field_menu.py
```diff
@@ -110,7 +110,9 @@ def __init__(self, sg_entity_type, parent=None, bg_task_manager=None, project_id
                 start_processing=True
             )
 
-            shotgun_globals.register_bg_task_manager(self._task_manager)
+            # removing this for now... these registrations were causing RV to crash on exit.
+            # are they necessary?
+            # shotgun_globals.register_bg_task_manager(self._task_manager)
```
robblau: We should have an answer to this before release.  Or if it is ok to live without the registration take it out.

python/shotgun_menus/entity_field_menu.py
```diff
@@ -164,6 +166,10 @@ def set_entity_type_filter(self, entity_type_filter):
         """
         self._entity_type_filter = entity_type_filter
 
+    # this was causing RV to segfault on exit. i think
+    # the task manager was already freed in another thread
+    # before this happens.
+    # removed the registration above and the crashing stopped.
```
robblau: Same as above.  We should either figure out the crash with the registration or remove it and this comment.

python/version_details/shotgun_entities/card_widget.py
```diff
@@ -350,6 +350,29 @@ def _set_entity(self, entity):
...
+
+        bundle = self.field_manager._task_manager._bundle
+        sg_res = bundle.shotgun.update(entity['type'], entity['id'], update_dict)
+
+        # TODO: what does sg_res look like in error?
+
+        # print "UPDATE: %r %r updated as follows:" % ( sg_res['type'], sg_res['id'] )
+        # for k in sg_res:
+        #     if k  != 'type' and k != 'id':
+        #         print "    %r is now %r" % ( k, sg_res[k])
```
robblau: Should this be cleaned up?

python/shotgun_menus/entity_field_menu.py
josh-t: I've been getting crashes on exit with `tk-multi-demos`. I think it only occurs when a demo is loaded that uses a task manager. I can double check and test what happens if one isn't registered.

josh-t:
Yep... I left myself a `TODO` and totally missed it here: https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/master/python/shotgun_fields/shotgun_field_editable.py#L23-L24

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -98,6 +98,13 @@ def _on_current_changed(self, index):
 
         self.currentWidget().setFocus()
 
+    def __getattr__(self, name):
+        """
+        Routes any attributes not found on the editable widget to the
+        fields widget that it is wrapping.
+        """
+        return getattr(self._display.display_widget, name)
```
josh-t: Maybe the methods do need to be handled individually. I think we need to forward `value_changed` signals from each of `editor` and `display`. 

`get_value` and `set_value` calls should probably update the values of the currently displayed widget. 

Would be nice to implement `enable_editing` to allow callers to programmatically swap the state. That method is required for widgets that are both display and edit so it would provide nice consistency.

python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -152,6 +159,13 @@ def eventFilter(self, obj, event):
 
         return False
 
+    def __getattr__(self, name):
```
josh-t: Since this one only wraps the display, maybe this is ok. I guess it should forward `value_changed` though.

python/shotgun_menus/entity_field_menu.py
thebeeland: I'm going to take it out for now. I don't see any adverse affects, and the crash on close no longer occurs.

PR #37, opened 2016-08-24
#37784 Fixes for entity edit widgets as per QA notes:

josh-t:
* added an item to the completer to indicate that the minimum number of
  characters to get results is 3. will now show "Type at least 3 characters..."
* completer popup will now be wide enough to show the complete results
    * smaller widgets were displaying popup that matched the width of the
      widget itself.
* added a method to manually clear the contents of the completer
* added a method to return the first result in the completer
    * this makes for it quick and easy to add entities to a multi entity widget
    * if no results are selected, the first will be used by hitting tab or enter
    * no more need to navigate to the first item to select it then hit enter
* the search widget now exposes the new entity_activated signal from the
    completer
    * connecting to this directly rather than having a subclass access the
        completer directly seems to address some of the reliability issues
    * not sure why
* minor bug fix to bubble widget
* turned off immediate reply for entity widget
    * this means the apply button (checkmark) is now visible again
    * this was done to facilitate clearing the contents to remove the value
      (which is valid)
* implemented get_value for entity widget
    * this will return None if the text is empty, otherwise the stored value
    * this makes the apply work for clearing the contents
* added key press short cuts for applying the value in the entity widget
    if the text has been cleared
* only show the completer in the multi entity widget focus in if its not
    already visible
    * this fixes the issue that was preventing clicks on popup
* updated "Known Issues" in docs

### DIFF
--- a/python/global_search_completer/global_search_completer.py
+++ b/python/global_search_completer/global_search_completer.py
@@ -44,9 +44,11 @@ class GlobalSearchCompleter(QtGui.QCompleter):
     SG_DATA_ROLE = QtCore.Qt.UserRole + 1
     MODE_ROLE = QtCore.Qt.UserRole + 2
 
+    COMPLETE_MINIMUM_CHARACTERS = 3
+
     # different items in the auto complete list can have
     # a different meaning, so track those here too
-    (MODE_LOADING, MODE_NOT_FOUND, MODE_RESULT) = range(3)
+    (MODE_LOADING, MODE_NOT_FOUND, MODE_RESULT, MODE_NOT_ENOUGH_TEXT) = range(4)
 
     def __init__(self, parent=None):
         """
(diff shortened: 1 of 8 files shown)


python/global_search_widget/global_search_widget.py
```diff
@@ -25,6 +25,7 @@ class GlobalSearchWidget(QtGui.QLineEdit):
 
     # emitted when shotgun has been updated
     entity_selected = QtCore.Signal(str, int)
+    entity_activated = QtCore.Signal(str, int, str)
```
thebeeland: Signal docs.

python/global_search_completer/global_search_completer.py
```diff
@@ -265,7 +293,7 @@ def set_searchable_entity_types(self, types_dict):
     ############################################################################
     # internal methods
 
-    def _clear_model(self, add_loading_item=True):
+    def _clear_model(self, add_loading_item=True, add_more_text_item=False):
```
thebeeland: Param docstring.

PR #40, opened 2016-09-01
modifications to styling specifically for editing fields in RV.

twkstewartb:


python/shotgun_fields/shotgun_field_editable.py
```diff
@@ -299,7 +299,7 @@ def __init__(self, display_widget, parent=None):
         self._edit_btn.hide()
 
         # make sure there's never a bg color or border
-        self._edit_btn.setStyleSheet("background-color: none; border: none;")
+        self._edit_btn.setStyleSheet("background-color: none; border: none; min-width: 15px;")
```
jfboismenu: maybe introduce a button_style constant?

python/version_details/shotgun_entities/card_delegate.py
```diff
@@ -167,11 +171,8 @@ def _get_painter_widget(self, model_index, parent):
         if model_index in self._widget_cache:
             widget = self._widget_cache[model_index]
 
-            if self.fields != widget.fields:
-                widget.fields = self.fields
-                self.sizeHintChanged.emit(model_index)
-
-            return widget
+            if sorted(self.fields) == sorted(widget.fields):
```
jfboismenu: Not a big deal, but if we had made fields a set during __init__, then we wouldn't have to sort these every single time. If fields is read-only, then frozenset would make even more sense.

thebeeland: Yeah, I agree that this is unfortunate. However, ordering is important in most cases in this chain of widgets, os simply using a set to store the list of fields wouldn't be appropriate in other situations where that data is utilized.

python/shotgun_fields/shotgun_field_editable.py
thebeeland: Agreed!

PR #41, opened 2016-09-02
Do not show new note dialog if there is no current entity.

eliveikis:
https://jira.autodesk.com/browse/GAME-17158

In Dynamite, if the user started drawing with no version loaded, the new note panel was showing. It should only show if there is a version loaded to add a note to. This change early outs if that is the case.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -614,8 +614,8 @@ class ActivityStreamWidget(QtGui.QWidget):
         :param bool modal: Whether the dialog should be shown modally or not.
         """
         if self._entity_id == None:
-            self._bundle.log_warning("Skipping New Note Dialog - No entity loaded.")
-            return;
+            self._bundle.log_debug("Skipping New Note Dialog - No entity loaded.")
+            return
 
         note_dialog = NoteInputDialog(parent=self)
         note_dialog.entity_created.connect(self._on_entity_created)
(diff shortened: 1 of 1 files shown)


eliveikis:
@thebeeland  Please review

python/activity_stream/activity_stream.py
```diff
@@ -613,6 +613,10 @@ def show_new_note_dialog(self, modal=True):
 
         :param bool modal: Whether the dialog should be shown modally or not.
         """
+        if self._entity_id == None:
+            self._bundle.log_warning("Skipping New Note Dialog - No entity loaded.")
+            return;
```
thebeeland: Ditch the semicolon!

python/activity_stream/activity_stream.py
```diff
@@ -613,6 +613,10 @@ def show_new_note_dialog(self, modal=True):
 
         :param bool modal: Whether the dialog should be shown modally or not.
         """
+        if self._entity_id == None:
+            self._bundle.log_warning("Skipping New Note Dialog - No entity loaded.")
```
thebeeland: I would probably make this a log_debug instead of a warning. I would consider this to be information that an artist/lead/sup using the app would need to see front and center.

python/activity_stream/activity_stream.py
```diff
@@ -613,6 +613,10 @@ def show_new_note_dialog(self, modal=True):
 
         :param bool modal: Whether the dialog should be shown modally or not.
         """
+        if self._entity_id == None:
```
manneohrstrom: nitpick: `if self._entity_id is None:` is the idiom we use in the rest of the code base. See [PEP-8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

_Comparisons to singletons like None should always be done with is or is not , never the equality operators._

PR #42, opened 2016-09-09
#38326 Removes project-only workarounds for completer in entity/multi-entity fields

josh-t:
Now allows for project-only search completion with the SG side bug fix. Required adding a special case whereby the completer detects that the query is only for Project entities and does not restrict by the current project. All other scenarios will continue to restrict by the current project if it can be determined.

### DIFF
--- a/python/global_search_completer/global_search_completer.py
+++ b/python/global_search_completer/global_search_completer.py
@@ -341,7 +341,12 @@ class GlobalSearchCompleter(QtGui.QCompleter):
 
         # constrain by project in the search
         project_ids = []
-        if self._bundle.context.project:
+
+        if len(self._entity_search_criteria.keys()) == 1 and \
+           "Project" in self._entity_search_criteria:
+            # this is a Project-only search. don't restrict by the current project id
+            pass
+        elif self._bundle.context.project:
             project_ids.append(self._bundle.context.project["id"])
 
         # run the query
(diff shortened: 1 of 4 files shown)


josh-t:
Now I'm thinking this requires some backward compatibility... hmmm. what do you think?

manneohrstrom:
so am i right thinking that you would now get this:


```
# completer should show assets and shots belonging to the current project context
completer. set_searchable_entity_types({
            "Asset": [],
            "Shot": [],
          })

# now we also want it to search for projects

# but adding a project switches over the completer into "site" mode so now will show matches for assets and shots for all projects.

completer. set_searchable_entity_types({
            "Project": [],
            "Asset": [],
            "Shot": [],
          })
```

The above seems confusing to me! I *think* up to now, all search results have been project specific (right?). But now suddenly we are allowing for cross project searching.

http://developer.shotgunsoftware.com/tk-framework-qtwidgets/global_search_completer.html

I am not totally sure, but maybe we need to add a flag to allow a user to better control whether results returned should be for the current project or across all projects?

```
# assets and shots for current project
completer. set_searchable_entity_types({
            "Asset": [],
            "Shot": [],
          })

# assets, shots and project for the current project - kind of pointless because only project 
# returned is the current project
completer. set_searchable_entity_types({
            "Asset": [],
            "Shot": [],
            "Project": [],
          })

# assets and shots across all proejcts
completer.search_all_projects = True
completer. set_searchable_entity_types({
            "Asset": [],
            "Shot": [],
          })

# search all projects
completer.search_all_projects = True
completer. set_searchable_entity_types({
            "Project": [],
          })
```

something like that?

josh-t:
Actually, the project restriction is still there except for the case when *only* Project entities are being searched. I felt like that was a smaller jump to make and probably accounts for most of the use cases. 

I actually started adding in a flag to control whether the completer was constrained by the current project ID, but I took it out. Let me know if you still think it is warranted. :)

PR #43, opened 2016-09-16
Game #16288: Attachments download refactor.

thebeeland:
Rather than relying on the activity stream cache being up to date with the current list of Attachment entities for a Note, we now query for them live and download what we find asynchronously.

### DIFF
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -479,8 +479,7 @@ class VersionDetailsWidget(QtGui.QWidget):
     def download_note_attachments(self, note_id):
         """
         Triggers the attachments linked to the given Note entity to
-        be downloaded. When the download is completed successfully, an
-        attachment_downloaded signal is emitted.
+        be downloaded.
 
         :param int note_id: The Note entity id.
         """
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -1235,13 +1235,23 @@ class VersionDetailsWidget(QtGui.QWidget):
         Field filter method for the EntityFieldMenu. Determines whether the
         given field should be included in the field menu.
 
-        :param field:   The field name being processed.
+        :param str field: The field name being processed.
         """
-        # Allow any fields that we have a widget available for.
-        return bool(
-            self.ui.current_version_card.field_manager.supported_fields(
-                "Version",
-                [field],
-            )
-        )
+
+        # see if we can display this field
+        supported = self.ui.current_version_card.field_manager.supported_fields(
+            "Version", [field])
+
+        if field not in supported:
+            return False
+
+        # get the current version entity's project id
+        entity = self.current_entity or {}
+        project_id = entity.get("project", {}).get("id")
+
+        # make sure the field is visible
+        if not shotgun_globals.field_is_visible("Version", field, project_id=project_id):
+            return False
+
+        return True
 
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -121,6 +121,7 @@ class VersionDetailsWidget(QtGui.QWidget):
         self._note_fields = [self.NOTE_METADATA_FIELD]
         self._attachments_filter = None
         self._dock_widget = None
+        self._pre_submit_callback = None
 
         self.ui = Ui_VersionDetailsWidget() 
         self.ui.setupUi(self)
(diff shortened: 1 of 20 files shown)


python/version_details/version_details.py
```diff
@@ -466,11 +467,25 @@ def download_note_attachments(self, note_id):
 
         :param int note_id: The Note entity id.
         """
-        attachments = self.get_note_attachments(note_id)
-
-        for attachment in attachments:
-            uid = self._data_retriever.request_attachment(attachment)
-            self._attachment_uids[uid] = note_id
+        # We're going to query the list of attachments live, because don't
```
josh-t: minor: because **we** don't

python/version_details/version_details.py
```diff
@@ -738,6 +751,15 @@ def _load_stylesheet(self):
...
             f.close()
 
+    def _download_attachments(self, attachments, note_id):
+        """
+        Downloads the contents of the given list of Attachment entities.
+
+        :param list attachments: A list of Attachment entities to download.
+        """
+        for attachment in attachments:
+            self._attachment_uids[self._data_retriever.request_attachment(attachment)] = note_id
```
josh-t: My preference would be to break this into 2 lines for readability.

python/version_details/version_details.py
```diff
@@ -116,6 +116,7 @@ def __init__(self, bg_task_manager, parent=None, entity=None):
         self._note_metadata_uids = []
         self._note_set_metadata_uids = []
         self._uploads_uids = []
+        self._attachment_query_uids = {}
```
josh-t: Looks like `_upload_uids` is no longer in use. Remove the line above?

python/version_details/version_details.py
```diff
@@ -738,6 +751,15 @@ def _load_stylesheet(self):
         finally:
             f.close()
 
+    def _download_attachments(self, attachments, note_id):
+        """
+        Downloads the contents of the given list of Attachment entities.
+
+        :param list attachments: A list of Attachment entities to download.
```
josh-t: missing param docs for `note_id`

python/version_details/version_details.py
```diff
@@ -466,11 +467,25 @@ def download_note_attachments(self, note_id):
...
-        for attachment in attachments:
-            uid = self._data_retriever.request_attachment(attachment)
-            self._attachment_uids[uid] = note_id
+        # We're going to query the list of attachments live, because don't
+        # know if the cached data for the activity stream is up to date. The
+        # reason that might be the case is that a new attachment doesn't
+        # trigger a new activity event, so the cache doesn't know it's out
+        # of date in that regard. It would be great to find a better solution
+        # than not trusting the cache.
+        self._attachment_query_uids[self._data_retriever.execute_find(
```
josh-t: My preference would be to break this up into 2 statements for readability. 13 line assignment is a bit long :P

manneohrstrom: +1 :P

python/version_details/version_details.py
```diff
@@ -484,11 +485,25 @@ def download_note_attachments(self, note_id):
...
+            "Attachment",
+            [[
+                "attachment_links",
```
manneohrstrom: docstring says a signal is emitted - is that still the case?

python/version_details/version_details.py
```diff
@@ -484,11 +485,25 @@ def download_note_attachments(self, note_id):
...
+                "in",
+                {"type":"Note", "id":note_id}
+            ]],
+            fields=[
+                "this_file",
+                "image",
```
manneohrstrom: How does this handle S3 urls with expiry? Is that handled elsewhere? How was it working before? Might be worth a comment.

python/version_details/version_details.py
thebeeland: Just relax your eyes. It's like a magic-eye picture.

python/version_details/version_details.py
thebeeland: Unless those expire nearly instantly it should never be a problem. Since we're querying the list of attachment entities live each time the download is requested we should be getting valid S3 urls.

python/version_details/version_details.py
gomezll: fixed

python/version_details/version_details.py
gomezll: broken up into 2 lines as requested

python/version_details/version_details.py
gomezll: Added

python/version_details/version_details.py
gomezll: broken up

python/version_details/version_details.py
gomezll: Doesn't seem like it is emitted. @thebeeland Can you verify this for us?

PR #47, opened 2016-09-22
Tweaks to NoteInputDialog

manneohrstrom:
Improved the docs for the `NoteInputDialog`.
Changed the constructor to not be open but rather specific in order to make it easier to handle future changes to the interface.

### DIFF
--- a/python/note_input_widget/dialog.py
+++ b/python/note_input_widget/dialog.py
@@ -14,6 +14,20 @@ from . import NoteInputWidget
 class NoteInputDialog(QtGui.QDialog):
     """
     A dialog wrapper for the :class:`NoteInputWidget` widget.
+
+    The dialog instance will mirror all attributes and methods
+    found on the embedded :class:`NoteInputWidget` instance.
+    Example::
+
+        note_dialog = NoteInputDialog(parent=self)
+        note_dialog.entity_created.connect(self._on_entity_created)
+        note_dialog.data_updated.connect(self.rescan)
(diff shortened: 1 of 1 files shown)


python/note_input_widget/dialog.py
```diff
@@ -14,12 +14,27 @@
...
+        note_dialog.data_updated.connect(self.rescan)
+        note_dialog.set_bg_task_manager(self._task_manager)
+        note_dialog.set_current_entity(self._entity_type, self._entity_id)
+
+        # show modal
+        note_dialog.exec_()
+
     """
-    def __init__(self, *args, **kwargs):
+    def __init__(self, parent):
```
manneohrstrom: This is how we do it everywhere else in the code. I think in general, we should try to avoid the `*args, **kwargs` open syntax unless we have reason to use it. It can make it trickier to make changes in the future - say for example we wanted to add some parameters to this method in the future.

python/note_input_widget/dialog.py
```diff
@@ -14,12 +14,27 @@
...
         """
-        Constructor.
```
manneohrstrom: minor: this was looking odd in the sphinx docs!

manneohrstrom:
We have something similar here - do you think it would be worth trying to consolidate the two? 

https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/master/python/activity_stream/dialog_reply.py

PR #49, opened 2016-09-27
Version details destroy function fixes.

eliveikis:
Fixed misnamed variable and also call `destroy` on the `ActivityStreamWidget` so that it can stop its data manager.

This is dependent on https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/49 to work, but will not cause problems without it.


eliveikis:
@manneohrstrom Please review

PR #54, opened 2016-10-09
EntityCardWidget: Fixed deletion of list items while iterating.

eliveikis:
The `remove_field` function was deleting the element from the `_fields` list and it was being called from iterations of that list.

This change separates the destruction of the field and its deletion from the list.

### DIFF
--- a/python/version_details/shotgun_entities/card_widget.py
+++ b/python/version_details/shotgun_entities/card_widget.py
@@ -151,7 +151,8 @@ class ShotgunEntityCardWidget(QtGui.QWidget):
         field_names = self.fields
 
         for field_name in field_names:
-            self.remove_field(field_name)
+            self.destroy_field(field_name)
+        self._fields = _OrderedDict()
 
     def get_visible_fields(self):
         """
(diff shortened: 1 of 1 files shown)


eliveikis:
@manneohrstrom Please review. There were other changes made in September that also combined to fix https://jira.autodesk.com/browse/GAME-17417, this is that last remaining piece.

python/version_details/shotgun_entities/card_widget.py
```diff
@@ -186,20 +187,27 @@ def remove_field(self, field_name):
...
+        field_widget.deleteLater()
 
         # If there's a label, then also remove that.
         field_label = self._fields[field_name]["label"]
 
         if field_label:
             field_label.hide()
             self.ui.field_grid_layout.removeWidget(field_label)
+            field_label.setParent(None)
+            field_label.deleteLater()
```
eliveikis: @manneohrstrom, @thebeeland   Is this necessary? I saw this pattern used elsewhere in toolkit when removing widgets and we want to make sure things are cleaned up properly so I added this here and above.

eliveikis:
will push another version of this soon

gomezll:
@thebeeland , @manneohrstrom Are you guys OK with these changes? If so, approve for merge. Thanks!

python/version_details/shotgun_entities/card_widget.py
thebeeland: I suspect this pattern is to plug a memory leak. I don't have first-hand experience with it, but I think it's reasonable to use it here given the situation.

PR #55, opened 2016-10-18
Handle time now is less than created time for widget time label.

eliveikis:
Remove early out from `_set_timestamp` so that it always sets the timestamp. If the created date is in the future, then use time now.

Nothing uses the return value of `_set_timestamp` so this should be a safe change.

### DIFF
--- a/python/activity_stream/widget_activity_stream_base.py
+++ b/python/activity_stream/widget_activity_stream_base.py
@@ -115,20 +115,21 @@ class ActivityStreamBaseWidget(QtGui.QWidget):
         
         # standard format 
         full_time_str = datetime_obj.strftime('%a %d %b %Y %H:%M') 
-    
-        if datetime_obj > datetime.datetime.now():
-            # future times are reported precisely
-            return (full_time_str, full_time_str) 
-        
+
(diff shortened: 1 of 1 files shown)


python/activity_stream/widget_activity_stream_base.py
```diff
@@ -115,20 +115,21 @@ def _set_timestamp(self, data, label):
...
         full_time_str = datetime_obj.strftime('%a %d %b %Y %H:%M') 
-    
-        if datetime_obj > datetime.datetime.now():
-            # future times are reported precisely
-            return (full_time_str, full_time_str) 
-        
+
+        time_now_obj = datetime.datetime.now()
+        # The note created time is rounded up to the nearest second by Shotgun,
+        # so handle creation time greater than now.
```
eliveikis: I'm not sure this is true in all cases, but it was observed to be true for "Reply" entity types. Note Reply creation dates can be rounded to the next minute. I'm not sure if this happens all the time and it depends on how long it takes for the panel to refresh, or if it only happens sometimes, since this bug was not visible all the time for us.

manneohrstrom:
CR LGTM! :shipit:

PR #57, opened 2016-10-19
Use underMouse instead of rect contains to detect note selection.

eliveikis:
The following reproduction steps revealed a strange bug with note selection in Dynamite:
1. Create two notes
2. Select the bottom note and add a reply
3. Select the other note (now at the bottom) and add a reply
Click back and forth in the notes and multiple times in the following areas:
image

Note that the notes will be improperly deselected or selected.

This problem was fixed by using the `underMouse` property instead of checking the rect.


eliveikis:
@thebeeland  Please review.

PR #58, opened 2016-10-20
Fix schema cache loading errors

eliveikis:
When using the Version Details panel for the first time with a clean Appdata folder, the project schema was not being cached early enough. This would lead to bugs like the Version card widget failing to show the fields, and also the remainder of the version refresh function callback would fail so things like `self.current_entity` would never be set, causing failures later on for anything relying on that.

This change adds a trigger for the schema load as early as possible (when the entity is first requested), and then moves all version finalization code to happen after the cache has been ensured to have been loaded.


eliveikis:
@manneohrstrom @robblau or @thebeeland  Please review

python/version_details/version_details.py
```diff
@@ -816,13 +822,11 @@ def _version_entity_data_refreshed(self):
         )
 
         if not item:
+            sgtk.platform.current_bundle().log_error("Unable to setup version summary: no entity item from version model.")
```
gomezll: @manneohrstrom  I've addressed this knitpick

gomezll:
@manneohrstrom, @thebeeland  I've addressed the knitpick you had above Manne. Looping you in @eliveikis Are you guys OK with the rest of these changes? Either way let me know, and if all good I'll add this code review to my list of merges planned for the end of this week. Thanks.

gomezll:
Oh and i also merged into this branch the latest changes from master to prep for final push.

eliveikis:
@gomezll Latest changes look good to me but need approval from @thebeeland

python/version_details/version_details.py
```diff
@@ -520,6 +520,10 @@ def load_data(self, entity):
             self.clear()
             return
 
+        # Ensure project schema will be loaded. The project schema is needed
+        # to reference Version fields in various UI and other elements.
+        shotgun_globals.run_on_schema_loaded(lambda : None, entity.get("project"))
```
thebeeland: I'm curious, as well. I believe that run_on_schema_loaded is asynchronous, so this isn't going to stop anything below from running. The lambda will be called, but it's not real clear to me what it's actually going to do.

gomezll: If you look at the run_on_schema_loaded implementation

```
  def run_on_schema_loaded(cls, callback, project_id=None):
        Run the given callback once the schema is loaded.

        :param callback:    Method with no argument to run when the schema is loaded
        :param project_id:  The id of the project entity to load the schema for. If
                            None, the current context's project will be used.

        self = cls.__get_instance()

        if self._is_schema_loaded(project_id=project_id):
            callback()
        else:
            self.schema_loaded.connect(callback)

            # kick off full schema loading
            self._check_schema_refresh(project_id=project_id)

```
You see that the code will kick off the schema loading in the else case. So this code isn't expecting to block or wait for the load to happen. It will just trigger the cache to be loaded sooner. Which will help make the data available to the UI sooner, before it is populated. The lambda does nothing probably because @eliveikis was interested in the triggering the cache to be populated.

What's not so clear to me is there is code that will synchronize and guarantee that the UI will be populated always AFTER the cache has populated. Any insights into how that happens @eliveikis?

eliveikis: @thebeeland  and @gomezll , Great points. Yeah it's been several months now so I'm not sure exactly the reasoning for this specific function call.
The odd thing is if I remember correctly, the bug only occurred when there is *no cache* so I'm surprised this would have fixed it if the cache isn't available yet so it should be async.
@gomezll I will send you the Jira with the reproduction steps and review the code a bit more.

eliveikis: @thebeeland and @gomezll Ok so it's clear now. So the particular call this comment is on, line 527 in `load_data`, is really just to make sure the schema is requested to be loaded as early as possible when a new entity is requested, for that entity's project. That is why lambda none is used since don't actually need to do anything yet.

The more critical change and actual fix to the bug is in the `_version_entity_data_refreshed` function. Before, that function would do some calls in the middle and below that relied on the schema cache being loaded, if not an exception was through and the rest of the entity setup would be skipped.

That function was changed to remove those bits and defer them to the new `_finalize_requested_version` function. 99% of the time, the schema is loaded at this point and the function is called right away, but rarely (in our case when launching from web only with no previous cache available), it is not.

gomezll: Thanks for explaining @eliveikis makes sense to me now and seems correct.

@thebeeland Let us know if you are good with @eliveikis explanation and all your concerns are addressed and give a final approval for merging. Thanks

eliveikis:
@gomezll Can you try removing line 527 and test launch from web from shotgun site with a fresh (empty) appdata? It might not be needed as @thebeeland mentioned. Maybe would be good to put a log there to see if the schema refresh request has already been kicked off or not.

PR #59, opened 2016-11-16
 #39094 Field menu polish

josh-t:
* Field menu uses current project id if available when querying the schema
* Styling to more closely match web menus
* shotgun menu groups can now be exclusive (only one item checkable)


python/shotgun_menus/entity_field_menu.py
```diff
@@ -353,3 +358,22 @@ def _get_qaction(self, field, display_name):
...
+    def _get_current_project_id(self):
+        """
+        Return the id of the current project.
+
+        :returns: The project id associated with the current context, or ``None``
+            if operating in a site-level context.
+        :rtype: ``int`` or ``None``
+        """
+
+        if self._bundle.tank.pipeline_configuration.is_site_configuration():
```
manneohrstrom: this is not part of the official core API.

python/shotgun_menus/entity_field_menu.py
```diff
@@ -353,3 +358,22 @@ def _get_qaction(self, field, display_name):
...
+            # site configuration (no project id). Return None which is
+            # consistent with core.
+            project_id = None
+        else:
+            project_id = self._bundle.tank.pipeline_configuration.get_project_id()
```
manneohrstrom: pipeline_configuration is not part of the official core API.

PR #61, opened 2016-12-04
 #38753 Makes it possible to use field delegate without SG model

josh-t:
Adds a data_role property to the delegate that defaults to the current data role from the SG model but allows you to set it to anything. These changes maintain the current behavior but also account for usage of non-SG models. Because SG models do some special things with the first column in the model and the delegate can use that for efficiency sake, the delegate will now inspect the indexes being processed to determine if the source model is a SG model or not and behave accordingly.

I'm not 100% sure this is the best solution here, but it seems to work. I've created a demo for this in the demo app that shows its usage.

https://github.com/shotgunsoftware/tk-multi-demo/blob/master/python/tk_multi_demo/demos/field_widget_delegate/demo.py#L109-L151

Will be great to get some other eyes on this. It doesn't feel ideal to me so I'm anxious to know if someone has a better approach.

### DIFF
--- a/python/shotgun_fields/shotgun_field_delegate.py
+++ b/python/shotgun_fields/shotgun_field_delegate.py
@@ -58,9 +58,18 @@ class ShotgunFieldDelegate(views.WidgetDelegate):
     def data_role(self):
         """
         The item role used to get and set data associated with the fields being
-        represented by this delegate. The default role is:
+        represented by this delegate.
 
-            ``ShotgunModel.SG_ASSOCIATED_FIELD_ROLE``
+        The data in this role should be of a type that matches the SG field
+        the delegate instance represents. For example, if the delegate's
+        entity type and field are ``HumanUser.name``, then the data should be
+        a string.
+
+        One exception is fields of type ``image`` such as the thumbnail field
+        found on many entity types. This can be either a ``QtGui.QPixmap``
+        instance or a path to a file on disk.
+
+        The default role is ``ShotgunModel.SG_ASSOCIATED_FIELD_ROLE``.
         """
         return self._data_role
 
--- a/python/shotgun_fields/shotgun_field_delegate.py
+++ b/python/shotgun_fields/shotgun_field_delegate.py
@@ -20,6 +20,10 @@ class ShotgunFieldDelegate(views.WidgetDelegate):
     A delegate for a given type of Shotgun field.  This delegate is designed to work
     with indexes from a ShotgunModel where the value of the field is stored in the
     SG_ASSOCIATED_FIELD_ROLE role.
+
+    To use this delegate with other types of models, simply use the ``data_role``
+    property to identify the role in the source model where data should be set
+    and retrieved.
     """
     def __init__(self, sg_entity_type, field_name, display_class, editor_class, view, bg_task_manager=None):
         """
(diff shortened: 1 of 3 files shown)


python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -46,6 +50,30 @@ def __init__(self, sg_entity_type, field_name, display_class, editor_class, view
...
+        # by the user via the data_role property
+        self._data_role = shotgun_model.ShotgunModel.SG_ASSOCIATED_FIELD_ROLE
+
+    @property
+    def data_role(self):
+        """
+        The item role used to get and set data associated with the fields being
+        represented by this delegate. The default role is:
+
+            ``ShotgunModel.SG_ASSOCIATED_FIELD_ROLE``
```
robblau: Should document what the content of that role is expected to be.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -201,26 +231,28 @@ def setModelData(self, editor, model, index):
...
             return
 
         bundle = sgtk.platform.current_bundle()
 
-        # special case for image fields
-        if editor._field_name == "image":
+        # special case for image fields in the ShotgunModel. The SG model stores
+        # the image field in the first column. If the value has changed, set the
+        # icon value there.
+        if self._source_is_sg_model and editor._field_name == "image":
```
robblau: For non ShotgunModel is there a way to do images?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -327,30 +359,41 @@ def _forward_mouse_event(self, mouse_event, index):
...
 
-    src_index = _map_to_source(model_index)
-    if not src_index or not src_index.isValid():
-        # invalid index, do nothing
-        return
+        # special case for image fields in the ShotgunModel. The SG model has
+        # the ability to pre-query thumbnails for entities for efficiency. If
+        # this is the image field for an entity in the SG model, we can make use
+        # of the potentially pre-queried image available in the first column.
+        if self._source_is_sg_model(model_index) and widget._field_name == "image":
```
robblau: Same as before, is there a way to use images with a custom model?

josh-t:
screen shot 2016-12-05 at 3 54 22 pm

Screenshot of the demo app with images for both SG an non-SG model usage.

python/shotgun_fields/shotgun_field_delegate.py
sgsteph: Same comment here: is it possible to do images with something else than a ShotgunModel? And should this test moved into some method or property, like `is_image`? The exact same test is used somewhere else.

python/shotgun_fields/shotgun_field_delegate.py
sgsteph: Same comment for using it with something else than a ShotgunModel, and having a `is_image` property or method.

python/shotgun_fields/shotgun_field_delegate.py
josh-t: Yep, demo app has example of using images in non-SG model case. I will look at cleaning up the check!

josh-t:
We could expose multiple roles for edit/display on the delegate if you think that would be better. The field widgets themselves do the translation between the SG value and what is needed to display (and back), but if it makes sense to separate them for the non-SG model case then I don't have any problem with that.

Regarding using a separate delegate class specific to the SG model case, that was what I originally wanted to do. But since this is released now, we'd be introducing backward incompatibility. We could introduce a new class that handles the non-SG model case if you think that is preferable. We'd probably move some of the existing logic into a base class at that point. Let me know what you think. Definitely seems more future proof than what I have currently.

sgsteph:
I think it would be more future proof and cleaner to have a class dedicated for the ShotgunModel and another one not knowing anything about it, in two different .py files. May be name the latter one `ShotgunFieldBaseDelegate` if we want to keep backward compatibility with things already using `ShotgunFieldDelegate`?
Not sure if having an Edit and a Display role is really needed, I was just mentioning it as this seems to be Qt standards. Feel free to do whatever makes sense!

josh-t:
Closing this one in favor of a newer one: https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/64

PR #64, opened 2017-01-05
 #38573 Second pass to create a field delegate for use with non-SG model

josh-t:
* Adds a `ShotgunFieldDelegateGeneric` class that `ShotgunFieldDelegate` inherits from. The **Generic** class accepts a field data role to use for retrieving/setting data in the associated model. The `ShotgunFieldDelegate` class simply hard codes this to the role in the SG model.
* Added a manager method called `create_generic_delegate()` to make it easy for client code to create these delegates. The demo app example has been updated to use this new method:   https://github.com/shotgunsoftware/tk-multi-demo/commit/91468f
* fw docs were busted so this also includes a few tweaks to get rid of doc build errors outlined in internal ticket 40026
For reference, the original PR is here: https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/61

### DIFF
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -173,8 +173,9 @@ class ImageWidget(QtGui.QLabel):
             if (not self._delegate and
                 not isinstance(value, QtGui.QPixmap) and
                 not os.path.exists(value)):
-                # this is not a local file or a pre-created pixmap. we need to
-                # download it
+                # the widget isn't being used as a delegate (which never
+                # requires downloads) and the value is not a local file or a
+                # pre-created pixmap. so we need to download it.
                 self._needs_download = True
             self._display_value(value)
         self.value_changed.emit()
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -170,7 +170,9 @@ class ImageWidget(QtGui.QLabel):
         if value is None:
             self._display_default()
         else:
-            if not isinstance(value, QtGui.QPixmap) and not os.path.exists(value):
+            if (not self._delegate and
+                not isinstance(value, QtGui.QPixmap) and
+                not os.path.exists(value)):
                 # this is not a local file or a pre-created pixmap. we need to
                 # download it
                 self._needs_download = True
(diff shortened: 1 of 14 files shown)


python/version_details/version_details.py
```diff
@@ -322,7 +322,8 @@ def note_threads(self):
         The currently loaded Note threads keyed by Note entity id and
         containing a list of Shotgun entity dictionaries.
 
-        Example structure containing a single Note entity:
+        Example structure containing a single Note entity::
```
sgsteph: Just being curious: why an additional `:` here?

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -12,16 +12,30 @@
...
     """
-    A delegate for a given type of Shotgun field.  This delegate is designed to work
-    with indexes from a ShotgunModel where the value of the field is stored in the
-    SG_ASSOCIATED_FIELD_ROLE role.
+    A generic, model-agnostic, shotgun field widget delegate.
+
+    This class is designed to be used with any model that represents data that
+    can be stored in Shotgun fields.
+
+    The included ``ShotgunFieldDelegate`` class is designed to work specifically
```
josh-t: Yeah, I thought about that. It does feel a little weird, but since they're in the same file I wanted to make sure anyone reading the docs for the base class knew about the subclass as well in case they were looking for the SG model delegate.

python/version_details/version_details.py
josh-t: With the single colon it was throwing an error on sphinx build, I think because of the following lines being indented. The double colons triggers a code block I think. Made the error go away at least. :)

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -12,16 +12,30 @@
 from sgtk.platform.qt import QtCore, QtGui
 
 views = sgtk.platform.current_bundle().import_module("views")
-shotgun_globals = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_globals")
-shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
 
-class ShotgunFieldDelegate(views.WidgetDelegate):
+shotgun_globals = sgtk.platform.import_framework(
```
josh-t: Fixed for both `import_framework` calls.

python/shotgun_fields/image_widget.py
```diff
@@ -170,7 +170,9 @@ def set_value(self, value):
         if value is None:
             self._display_default()
         else:
-            if not isinstance(value, QtGui.QPixmap) and not os.path.exists(value):
+            if (not self._delegate and
```
josh-t: updated!

PR #66, opened 2017-01-10
Adding property to query the entity that is being loaded or was loade…

gomezll:
…d by the panel, if any. This is used in the fix to GAME-19131.

This is property is used in this fix to GAME-19131 in the pull request: https://github.com/shotgunsoftware/tk-multi-versiondetails/pull/18/files

### DIFF
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -304,9 +304,16 @@ class VersionDetailsWidget(QtGui.QWidget):
     @property
     def current_entity(self):
         """
-        The current Shotgun entity that is active in the widget.
+        The current Shotgun entity that has been loaded and is active in the widget.
         """
         return self._current_entity
+
+    @property
+    def active_entity(self):
+        """
+        The current Shotgun entity that is OR will become active in the widget.
+        """
+        return self._current_entity or self._requested_entity
 
     @property
     def is_pinned(self):


python/version_details/version_details.py
```diff
@@ -304,11 +304,18 @@ def destroy(self):
...
         """
-        The current Shotgun entity that is active in the widget.
+        The current Shotgun entity that has been loaded and is active in the widget.
         """
         return self._current_entity
 
     @property
+    def active_entity(self):
+        """
+        The current Shotgun entity that is OR will become active in the widget.
```
eliveikis: documentation missing some words

python/version_details/version_details.py
```diff
@@ -304,11 +304,18 @@ def destroy(self):
     @property
     def current_entity(self):
         """
-        The current Shotgun entity that is active in the widget.
+        The current Shotgun entity that has been loaded and is active in the widget.
         """
         return self._current_entity
 
     @property
+    def active_entity(self):
```
eliveikis: I'm wondering if having these two separate functions is too implementation specific. Is the fact that an entity was requested but has not finished loading from the web mean we should have two separate functions to retrieve the entity?   Should there just be a single `entity` function that behaves like the new `active_entity` instead?

Does the client of this code need to know if the entity is fully loaded? If so we could add a `is_entity_loaded` function that checks if `current_entity` is set.

gomezll: @eliveikis I have updated and simplified the code with your suggestion above. Please review changes in https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/71

PR #70, opened 2017-01-26
For #39721, add progress indicator to project bootstrap

jfboismenu:
Refactors the ShotgunOverlayWidget so that we can use the spinning part without the messaging part as the `ShotgunSpinnerWidget`. Also added the notion of progress reporting to the spinner.

This is what it looks like:
image
The dot at the end of the line is the heartbeat which provides visual indication that the UI is not frozen if it takes a long time between two updates.

### DIFF
--- a/python/overlay_widget/shotgun_spinning_widget.py
+++ b/python/overlay_widget/shotgun_spinning_widget.py
@@ -8,6 +8,9 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
+
+import math
+
 from tank.platform.qt import QtCore, QtGui
 
 # load resources
@@ -42,6 +45,7 @@ class ShotgunSpinningWidget(QtGui.QWidget):
         self._spin_angle = 0
         self._spin_angle_to = 0
         self._previous_spin_angle_to = 0
+        self._heartbeat = 0
 
         self._sg_icon = QtGui.QPixmap(":/tk_framework_qtwidgets.overlay_widget/sg_logo.png")
 
@@ -95,6 +99,8 @@ class ShotgunSpinningWidget(QtGui.QWidget):
             # If the current spin angle has not reached the destination yet,
             # increment it, but not past.
             self._spin_angle = min(self._spin_angle_to, self._spin_angle + 1)
+            self._heartbeat = (self._heartbeat + 1) % 25
+
         self.repaint()
 
     def _draw_opened_circle(self, painter, start_angle, span_angle):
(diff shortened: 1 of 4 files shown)


python/overlay_widget/shotgun_spinning_widget.py
```diff
@@ -0,0 +1,223 @@
...
+        self._timer.start(1000 / self._UPDATES_PER_SECOND)
+        self._mode = self.MODE_PROGRESS
+        self._spin_angle_to = 0
+        self._spin_angle = 0
+
+    def report_progress(self, current):
+        """
+        Updates the widget current progress value.
+
+        :param current: New value for the progress arc. Must be between 0 (nothing) and 1 (complete).
```
manneohrstrom: nitpick: maybe say between 0.0 and 1.0 instead to highlight the fact that they are floats.

python/overlay_widget/shotgun_spinning_widget.py
```diff
@@ -0,0 +1,223 @@
...
+
+        :param painter: Painter object we will draw with.
+        :param start_angle: Angle at which we will start drawing the arc.
+        :param span_angle: Degrees the arc covers.
+        """
+        # show the spinner
+        painter.translate((painter.device().width() / 2) - 40,
+                          (painter.device().height() / 2) - 40)
+
+        pen = QtGui.QPen(QtGui.QColor("#424141"))
```
manneohrstrom: can we use a core constant?

python/overlay_widget/shotgun_spinning_widget.py
```diff
@@ -0,0 +1,223 @@
...
+
+from tank.platform.qt import QtCore, QtGui
+
+# load resources
+from .ui import resources_rc # noqa
+
+
+class ShotgunSpinningWidget(QtGui.QWidget):
+    """
+    Overlay widget that can be placed on top over any QT widget.
```
manneohrstrom: did you check the sphinx build?

feels like we need some examples at the very least!

python/overlay_widget/shotgun_spinning_widget.py
jfboismenu: Not sure what you mean. If you mean using an already existing constant, none actually match. The color is the same as the icon its drawing around, so the color is more tied to that icon than any core values.

python/overlay_widget/shotgun_spinning_widget.py
jfboismenu: Hadn't thought about documenting with code samples.

manneohrstrom:
this has been merged so closing.

PR #71, opened 2017-01-31
Game 19131 same geometry path different notes loaded

gomezll:
Code review refinements to https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/66 suggested by @eliveikis

### DIFF
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -304,9 +304,16 @@ class VersionDetailsWidget(QtGui.QWidget):
     @property
     def current_entity(self):
         """
-        The current Shotgun entity that is active in the widget.
+        The current Shotgun entity that has been loaded and is active in the widget.
         """
         return self._current_entity
+
+    @property
+    def active_entity(self):
+        """
+        The current Shotgun entity that is OR will become active in the widget.
+        """
+        return self._current_entity or self._requested_entity
 
     @property
     def is_pinned(self):
--- a/python/version_details/version_details.py
+++ b/python/version_details/version_details.py
@@ -306,13 +306,6 @@ class VersionDetailsWidget(QtGui.QWidget):
         """
         The current Shotgun entity that has been loaded and is active in the widget.
         """
-        return self._current_entity
-
-    @property
-    def active_entity(self):
-        """
-        The current Shotgun entity that is OR will become active in the widget.
-        """
         return self._current_entity or self._requested_entity
 
     @property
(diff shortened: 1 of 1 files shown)


python/version_details/version_details.py
```diff
@@ -586,7 +587,7 @@ def load_data(self, entity):
         self._requested_entity = entity
 
         # If we're pinned, then we don't allow loading new entities.
-        if self._pinned and self.current_entity:
+        if self._pinned and self._current_entity:
```
eliveikis: This overall concept of the change looks good to me. I have a few concerns though:
1. Need to make sure that wherever the code was switched to use `self.current_entity`, that the code isn't dependent on some other data that doesn't exist yet until `self._current_entity` is set.
2. Need to go through other tk components that reference `version_details_panel.current_entity` and do the same - make sure that they aren't relying on some other data that is not available yet when `._request_entity` is set.
3. Need to go through and cleanup other tk components that have checks for `or panel._requested_entity` since I think I added several of those during the sg3dr work.

eliveikis: I can see you've done #3 in at least one other pr so that's good - would be good to see your thoughts on #2.

gomezll: Thanks for input @eliveikis. Yes, i've searched, checked, and made changes (where necessary) to all tk components currently used by dynamite as you saw in https://github.com/shotgunsoftware/tk-multi-versiondetails/pull/22 and pointed out above.

I presume you are referring to other tk components not in use by dynamite but that still have references to .current_entity and ._request_entity. Those I have not checked. So I will ask @robblau to please give me access to all other tk components, so i can check that any references in those won't break due to this change. Thanks for bringing that up.

gomezll: @eliveikis I've done a search thru github repositories for non-dynamite tk components @robblau  gave me access https://github.com/search?p=3&q=org%3Ashotgunsoftware+.current_entity+in%3Afile&type=Code (This search will not work if you don't have access to those repositories) And there was only one other instance where current_entity property was being queried and that seemed safe and should be unaffected by this change.

So I will make these changes now available to @thebeeland  for review.

gomezll: @eliveikis I've done a search thru github repositories for non-dynamite tk components @robblau  gave me access https://github.com/search?p=3&q=org%3Ashotgunsoftware+.current_entity+in%3Afile&type=Code (This search will not work if you don't have access to those repositories) And there was only one other instance where current_entity property was being queried and that seemed safe and should be unaffected by this change.

So I will make these changes now available to @thebeeland  for review.

gomezll: @thebeeland These changes will be merged into master. But requesting review on dynamite-lib here. Also, these changes go along with https://github.com/shotgunsoftware/tk-multi-versiondetails/pull/22

PR #72, opened 2017-02-17
Game 18568 fast note creation ownership

eliveikis:
SG3DR needs to immediately capture the application state when a note is created, not wait until the shotgun-id is available.

When a note is created by the application, this change adds the ability for the application to pass a userdata object along with the creation request that is returned in the callback for that note creation.

When a note is created by the toolkit, e.g. by the user using the `NoteInputWidget`, new signals have been added to immediately notify the application that a note is being create, and passing note creation task Ids with the creation request and when the note is finally created.

These two pathways allow the application to synchronize and match *requests* for new notes with the *creation confirmation callbacks*.

This fixes bugs when rapidly creating notes where the application state would change and shotgun notes would become queued and take a long time to create, resulting in state mismatch of when the note was originally created.

### DIFF
--- a/python/note_input_widget/widget.py
+++ b/python/note_input_widget/widget.py
@@ -620,7 +620,7 @@ class NoteInputWidget(QtGui.QWidget):
             self.clear()
             self._bundle.log_debug("Update call complete! Return data: %s" % data)
             self.data_updated.emit()
-            self.entity_created.emit(data["return_value"], uid)
+            self.entity_created.emit(data["return_value"], int(uid))
             if self._outgoing_tasks and self._outgoing_tasks.has(uid):
                 self._outgoing_tasks.remove(uid)
             self.__overlay.hide()
--- a/python/note_input_widget/widget.py
+++ b/python/note_input_widget/widget.py
@@ -33,6 +33,9 @@ class NoteInputWidget(QtGui.QWidget):
         replied to. 
     :signal close_clicked: Emitted if a user chooses to cancel the note 
         creation by clicking the X button.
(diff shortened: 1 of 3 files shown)


eliveikis:
@gomezll  Please review, needs cleanup and I'll point out an issue in the comments...
This enables the changes in:
https://github.com/shotgunsoftware/tk-multi-versiondetails/pull/23

python/note_input_widget/widget.py
```diff
@@ -46,14 +49,18 @@ class NoteInputWidget(QtGui.QWidget):
     data_updated = QtCore.Signal()
     close_clicked = QtCore.Signal()
 
+    # Emitted when a new note or reply will be created. The creation request id
+    # is sent and will match the id sent with the entity_created signal.
+    # Todo: This has functional overlap with the pre_submit_callback...resolve that...
+    new_entity_requested = QtCore.Signal(str, int)
```
eliveikis: @gomezll  I added these extra signals and functionality before I realized there is already a mechanism that supplies similar functionality - the `pre_submit_callback`. Imo the existing callback is a bit odd since it passes the `NoteInputWidget` object as its parameter which feels a bit arbitrary, and some work would be needed to use that one since the pending request UID would need to be exposed through a function and that also seems a bit random.

I took the approach of adding new signals specific for the exact event type that we're trying to track (shotgun note creations and external user creations (in activity_stream.py), which seems nice and explicit but also makes things a bit busier.

Interested to see what you think.

gomezll: Your approach with new signals seems justified and appropriate to me Ed. Looks good

PR #76, opened 2017-04-20
For #39978, hierarchy based completer

jfboismenu:
Introduces a search widget and completer based on the nav api from Shotgun. A lot of work has been done in order to refactor the existing global search widget code so it could be reused by this new search widget.

The base widget class introduces two new UX enhancements that weren't present in the original global search widget and which is now available to both search widgets:
- widget waits 300ms after the used stops typing before issuing a request to the server
- search results have the search term underlined
- clear button at the right end of the search box

The framework remains backwards compatible with the previous version of the API, but the official API has consolidated the hierarchical and global completers under the `search_completer` module and the Shotgun-enabled search widgets into `shotgun_search_widget` (`search_widget` was already used, which btw the `shotgun_search_widget` should probably be refactored in the future in order to use the `SearchWidget`)

Obviously Github makes it seems like files were deleted and created when in fact they were moved and slightly edited. I'll identify the sections of the code than have actually changed in those files.

Here's what the new hierarchical search looks like:
screen shot 2017-04-20 at 00 07 06

Here's what the global search looks like:
screen shot 2017-04-20 at 00 07 36

Here's the documentation for the two new public classes:
screen shot 2017-04-21 at 08 44 18
screen shot 2017-04-21 at 08 44 37


docs/search_completer.rst
```diff
@@ -0,0 +1,35 @@
+Shotgun Search Completers
+#############################################
+
+
+.. currentmodule:: search_completer
```
jfboismenu: This is one of the moved files, but I recommend actually re-reading it.

docs/shotgun_search_widget.rst
```diff
@@ -0,0 +1,20 @@
+Shotgun Search Widget
+#############################################
+
+The Shotgun search widgets are similar to the global search and hierarchical
```
jfboismenu: This is one of the moved files, but I recommend actually re-reading it.

info.yml
```diff
@@ -27,5 +27,5 @@ requires_core_version: "v0.14.58"
  
 # the frameworks required to run this app
 frameworks:
-    - {"name": "tk-framework-shotgunutils", "version": "v4.x.x"}
+    - {"name": "tk-framework-shotgunutils", "version": "v5.x.x"}
```
jfboismenu: Required in order to access newer methods from the ShotgunDataRetriever.

python/search_completer/global_search_completer.py
```diff
@@ -0,0 +1,219 @@
+# Copyright (c) 2016 Shotgun Software Inc.
```
jfboismenu: This isn't a new file. I'll point out what has changed.

python/search_completer/global_search_completer.py
```diff
@@ -0,0 +1,219 @@
...
+            "Version": [],
+            "PublishedFile": [],
+          }
+
+        :param types_dict: A dictionary of searchable types with optional filters
+
+        """
+        self._entity_search_criteria = types_dict
+
+    def _set_item_delegate(self, popup, text):
```
jfboismenu: These methods are new. These are part of the contract with the base class.

jfboismenu: These code under this line is new. Those methods are called by the base class, SearchCompleter

python/search_completer/global_search_result_delegate.py
```diff
@@ -0,0 +1,83 @@
...
+
+class GlobalSearchResultDelegate(SearchResultDelegate):
+    """
+    Delegate which renders search match entries in the global
+    search completer.
+    """
+
+    def _render_result(self, widget, model_index):
+
+        from .global_search_completer import GlobalSearchCompleter
```
jfboismenu: Most of the code from this class has been moved to SearchResultDelegate. The code itself hasn't changed, expect that it now underlines the search terms in the result.

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,187 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
```
jfboismenu: This is all new code.

python/search_completer/hierarchical_search_result_delegate.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
```
jfboismenu: This is new code.

python/search_completer/search_result_delegate.py
```diff
@@ -70,77 +71,47 @@ def _on_before_paint(self, widget, model_index, style_options):
...
-                else:
-                    # present type name name
-                    link_entity_type = links[0]
-                    content += " on %s %s" % (shotgun_globals.get_type_display_name(link_entity_type), links[1])
-            
-            widget.set_text(content)
-        
+
+        elif mode == SearchCompleter.MODE_RESULT:
+            self._render_result(widget, model_index)
```
jfboismenu: This is where we delegate the rendering of the result to the derived class.

python/search_completer/global_search_result_delegate.py
```diff
@@ -0,0 +1,93 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk.platform.qt import QtCore
+
+from .search_result_delegate import SearchResultDelegate
```
jfboismenu: Code from this class was taken from the original implementation of the `SearchResultDelegate`. The code itself hasn't changed, expect that it now underlines the search terms in the result.

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,187 @@
...
+        if len(matches) == 0:
+            item = QtGui.QStandardItem("No matches found!")
+            item.setData(self.MODE_NOT_FOUND, self.MODE_ROLE)
+            self.model().appendRow(item)
+
+        # Payload looks like:
+        # [
+        #     {
+        #         "label": "bunny_020",
+        #         "incremental_path": [
```
manneohrstrom: `incremental_paths` (with an s)?

info.yml
manneohrstrom: So does this mean we are fully v5. in the dependency chain for sg-frameworks now? We had a separate ticket to fix this - can you close this out when this is merged in? 🍬 !

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,187 @@
...
+shotgun_model = sgtk.platform.import_framework("tk-framework-shotgunutils", "shotgun_model")
+
+
+class HierarchicalSearchCompleter(SearchCompleter):
+    """
+    A standalone :class:`PySide.QtGui.QCompleter` class for matching SG entities to typed text.
+
+    :signal: ``node_activated(str, int, str, str, list)`` - Fires when someone activates a
+        node inside the search results. The parameters are ``type``, ``id``, ``name``,
+        ``label path`` and ``incremental_paths``. If the node activated is not an entity,
```
manneohrstrom: the hierarchy API returns different `kinds` of things - wondering if there are usecases where this would be useful to have returned. it feels like the current signal is perfect for when an entity is clicked (the normal case) but if something else is clicked it's a bit of a black box.

How does this work practically? if there is a hierarchy node `Assets > Vehicle (sg asset type) > Ford`, can i search in the completer for `Vehicle`? 

If yes, i think it would be helpful to have a mode where the completer only returns entities, so a mode where we know that the `node_activated` signal always returns `type` and `id` that are never `None`. I think this is the normal use case.

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        Allows to change the root of the search.
+
+        :param dict entity: Entity to search under. If ``None``, the search will be done
+            at the site level. Note that only ``Project`` entities are supported at the moment.
+        """
+        if not entity:
+            self._search_root = "/"
+        else:
+            self._search_root = "/Project/%d" % entity.get("id")
```
manneohrstrom: @daigles has suggested that in order to keep the navigation API as flexible as possible, we should not do any string manipulation of these paths at all - think of them as black box tokens being sent down from the server (i think of them as guids). Instead, pass the entity into the new resolve method to get paths back and then we can use the first path that is returned. That way, the client side is completely agnostic and all the resolve logic stays on the server side! in case there is more than one value returned we should log it (i don't think it's a warning but useful for debugging at the very least)

info.yml
jfboismenu: I simply bumped the version number, I have to admit I didn't ensure everything was still working correctly. Francis should probably to a pass on the demo app to test all the widgets. But in theory, yes, this is the last package that depended on v4 as far as I can tell.

python/search_completer/hierarchical_search_completer.py
jfboismenu: I think having a mode would be better. Its simple to filter out entries, plus we can't tell what people would like to do with the widget. Maybe I can add this as a separate PR to simplify the review process?

python/search_completer/hierarchical_search_completer.py
jfboismenu: I totally agree, but currently the nav API is broken and can't resolve the path for a project entity, so I'm stuck doing this. I think we should ticket up the work on the web and tk side.

python/search_completer/hierarchical_search_completer.py
jfboismenu: Nope. Payload is really `incremental_path`.

jfboismenu:
New version of the code:
- adds an option (enabled by default) to only show entities in the search results.
- documents why we are hardcoding the path string
- turns setters into properties

python/search_completer/global_search_result_delegate.py
```diff
@@ -0,0 +1,93 @@
...
+        #  'project_id': 65,
+        #  'type': 'Version',
+        #  'id': 99}
+
+        entity_type_display_name = shotgun_globals.get_type_display_name(data["type"])
+
+        content = ""
+        et_url = shotgun_globals.get_entity_type_icon_url(data["type"])
+
+        highlighted_name = self._highlight_search_term(data["name"])
```
josh-t: should this be called `_underline_search_term()`? ;)

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,227 @@
...
+        #             "id": 5,
+        #             "type": "Sequence"
+        #         },
+        #         "project_id": 65
+        #     },
+        #     ...
+        # ]
+
+        # insert new data into model
+        for d in matches:
```
josh-t: can we change this to `for data in data_matches` or something a little more descriptive?

python/search_completer/hierarchical_search_completer.py
```diff
@@ -0,0 +1,227 @@
...
+
+            item = QtGui.QStandardItem(d["label"])
+            item.setData(self.MODE_RESULT, self.MODE_ROLE)
+
+            item.setData(shotgun_model.sanitize_for_qt_model(d),
+                         self.SG_DATA_ROLE)
+
+            item.setIcon(self._pixmaps.no_thumbnail)
+
+            if d["ref"]["type"] and d["ref"]["id"] and self._sg_data_retriever:
```
josh-t: I think this would read better if these were defined first before conditional:

```python
data_type = data["ref"]["type"]
data_id = data["ref"]["id"]

if data_type and data_id and self._sg_data_retriever:
    ...
    
    uid = self._sg_data_retriever.request_thumbnail_source(data_type, data_id, load_image=True)
```

python/search_completer/search_completer.py
```diff
@@ -313,51 +215,18 @@ def _clear_model(self, add_loading_item=True, add_more_text_item=False):
         self.model().clear()
 
         if add_loading_item:
-            item = QtGui.QStandardItem("Loading data...")
+            item = QtGui.QStandardItem("Hold on, loading search results...")
```
josh-t: I find **Hold on** a bit aggressive, implying that I'm being impatient. Maybe I'm just sensitive. Perhaps just **Loading search results...**?

python/search_completer/search_result_delegate.py
```diff
@@ -70,77 +71,52 @@ def _on_before_paint(self, widget, model_index, style_options):
...
+        if not self._text:
+            return matching
+
+        match_start = matching.lower().find(self._text.lower())
+        if match_start == -1:
+            return matching
+
+        match_end = match_start + len(self._text)
+
+        return "%s<span style='text-decoration:underline;'>%s</span>%s" % (
```
josh-t: slick!

python/search_widget/search_widget.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, parent=None):
         # set up the UI
         self._ui = Ui_SearchWidget()
         self._ui.setupUi(self)
-        self.set_placeholder_text("Search")
+        self.set_placeholder_text("Search...")
```
josh-t: Maybe this should be **Type here to search...**?

python/shotgun_search_widget/hierarchical_search_widget.py
```diff
@@ -0,0 +1,89 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+from sgtk.platform.qt import QtCore
+
+from .shotgun_search_widget import ShotgunSearchWidget
+
+search_completer = sgtk.platform.current_bundle().import_module(
```
josh-t: one line?

python/shotgun_search_widget/shotgun_search_widget.py
```diff
@@ -0,0 +1,126 @@
...
+        # are not the same, but at least the ShotgunSearchWidget has feature parity.
+        self.set_placeholder_text("Search")
+
+        # dynamically create the clear button so that we can place it over the
+        # edit widget:
+        self._clear_btn = QtGui.QPushButton(self)
+        self._clear_btn.setFocusPolicy(QtCore.Qt.StrongFocus)
+        self._clear_btn.setFlat(True)
+        self._clear_btn.setCursor(QtCore.Qt.ArrowCursor)
+        style = ("QPushButton {"
```
josh-t: Other `qtwidgets` use a `style.qss` file and load it. Any reason not to do that here? See: https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/bb0c5d2624d2b2df5190ae1d4b797eab4637ff64/python/note_input_widget/widget.py#L69

python/search_widget/search_widget.py
jfboismenu: Type here to search... might be too long for certain UIs. I'd rather stick with "Search..." for now.

python/shotgun_search_widget/shotgun_search_widget.py
jfboismenu: As the comment says above, I copy pasted this bit from somewhere else. I'll fix it.

PR #78, opened 2017-05-15
For #42310, search completer polish

jfboismenu:
Unfortunately this pull request doesn't cover as much ground as we wanted, but we need to cut our losses here and focus on what needs to absolutely needs to get in before the release, so this will do it for now.

Hitting ESC will clear the line edit and dismiss the popup, while picking a result will also clear the line edit.

I also remade some of the code Python 2.5 compatible to reduce the amount of risk for the release.

### DIFF
--- a/python/search_completer/search_result_delegate.py
+++ b/python/search_completer/search_result_delegate.py
@@ -1,16 +1,15 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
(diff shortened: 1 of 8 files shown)


python/search_completer/search_result_delegate.py
```diff
@@ -22,59 +21,71 @@
 from .utils import CompleterPixmaps
 
 
-class SearchResultDelegate(views.EditSelectedWidgetDelegate):
+class SearchResultDelegate(views.WidgetDelegate):
```
jfboismenu: Fixes some rendering issues with the completer.

Sometimes the first result would be gray rectangle, which turned out to because the widget was created in editor mode and cause some weird Qt-issues. Having the widget be a regular widget delegate fixed all of this. Becase of this however we now need to listen for selection events.

PR #80, opened 2017-08-17
For #40555: Only calls deleteLater of Qt objects when in Qt4.

thebeeland:
We're seeing a lot of problems related to garbage collection in PySide2/Qt5 when deleteLater is used. These issues result in crashes on DCC close in some cases, but also errors on app close even when the DCC isn't being closed.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -729,8 +729,14 @@ class ActivityStreamWidget(QtGui.QWidget):
                 self.ui.activity_stream_layout.removeWidget(x)
                 # set it's parent to None so that it is removed from the widget hierarchy
                 x.setParent(None)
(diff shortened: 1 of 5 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -729,8 +730,7 @@ def _clear(self):
                 self.ui.activity_stream_layout.removeWidget(x)
                 # set it's parent to None so that it is removed from the widget hierarchy
                 x.setParent(None)
-                # mark it to be deleted when event processing returns to the main loop
-                x.deleteLater()
+                utils.safe_delete_later(x(
```
josh-t: typo?

thebeeland: Fuck.

PR #81, opened 2017-10-30
fixed issue with thumbnails not properly downloading

mbrainerd:
It appears that the argument ordering is incorrect.


mbrainerd:
Any update on this request?

PR #82, opened 2017-12-04
For SPLAT-103: Add "Created Note" and "Created Reply" metrics

nmontmarquette:
Basically add base code for handling the "Created Note" and "Created Reply" metrics.
Further changes might populate some of the currently non-populated metric fields.


python/activity_stream/activity_stream.py
```diff
@@ -14,6 +14,8 @@
 
 from sgtk.platform.qt import QtCore, QtGui
 
+from . import utils as qtwidgets_utils
```
josh-t: is this needed?

python/activity_stream/metric_utils.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) 2017 Shotgun Software Inc.
```
josh-t: So in the other PRs, the property gathering and metric logging was inline with the rest of the code. Here you've moved the actual logging to a separate util file. Suggest we just do this inline to be consistent unless there's a strong reason not to.

python/activity_stream/reply_list.py
```diff
@@ -418,6 +420,7 @@ def _on_reply_clicked(self, note_id):
             self.__small_overlay.show()
             if reply_dialog.exec_() == QtGui.QDialog.Accepted:
                 self._data_manager.rescan()
+                log_metric_created_reply(self._bundle, "Reply List")
```
josh-t: Should this be done in the `ReplyDialog`?

python/activity_stream/metric_utils.py
nmontmarquette: The reason was that I needed to emit the same metric from different modules (2 for now) maybe more is the future?

python/activity_stream/reply_list.py
nmontmarquette: Would that be another one then? There is a field for source. One would be "Reply List" the one your are mentionning would be "Reply Dialog" ?

nmontmarquette: How do I know I'm in "Reply List" from within the QT Dialog class ?

PR #83, opened 2018-01-05
#46026 adds context widget to qt widgets fw

manneohrstrom:
Adds the context widget from the publisher to the fw. Minor cleanup and docs - the code was already in a very nice state. Docs look like this:

image


python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        return False
+
+    def save_recent_contexts(self):
+        """
+        Should be called by the parent widget, typically when the dialog closes,
+        to ensure the recent contexts are saved to disk when closing.
+        """
+
+        # build a list of serialized recent contexts. we grab all the QActions
+        # from the recents list ad serialize them.
```
josh-t: `and`

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+
+        self._context = context
+        self._show_context(
+            context,
+            task_display_override=task_display_override,
+            link_display_override=link_display_override
+        )
+
+        # ensure the new context is added to the list of recents.
+        # TODO: we should only update recents after publishing. but until we
```
josh-t: maybe remove this comment since this is outside the publisher now?

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        # clicked, they go directly to an edit state
+        self.ui.task_display.installEventFilter(self)
+        self.ui.link_display.installEventFilter(self)
+
+        # setup event filters for the task/link search inputs so that when
+        # certain keys are pressed, the widget can react to it properly
+        self.ui.task_search.installEventFilter(self)
+        self.ui.link_search.installEventFilter(self)
+
+        # we need to limit the search completer to entity types that are valid
```
josh-t: Should we reconsider this logic since this is no longer limited to the publisher?

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+
+        self.context_label.setText(message or "")
+
+    def _add_to_recents(self, context):
+        """
+        Adds the supplied context as an action in the list of recents context
+        actions
+        """
+
+        # don't add a "project" context to recents. we shouldn't encourage it
```
josh-t: publisher specific?

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        """
+        Build a list of actions from the supplied tasks. The actions are stored
+        in the instance's _menu_actions dictionary and used to build the menu.
+
+        The actions will be sorted by name if ``sort`` is set to True. If the
+        ``exclude_current_context`` is supplied, the widget's current context
+        will not be included in the list of actions.
+        """
+
+        publisher = sgtk.platform.current_bundle()
```
josh-t: maybe we define `self._bundle` in `__init__()` instead? we should rename this at least.

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        """
+        Slot called just before the contexts menu is shown. It handles
+        organizing the actions into menus.
+        """
+
+        # clear and rebuild the menu since the recents/related sections are
+        # dynamic.
+        self._task_menu.clear()
+
+        publisher = sgtk.platform.current_bundle()
```
josh-t: same comment about `self._bundle`.

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        # update the widget to display the new context and alert listeners that
+        # a new context was selected
+        self._show_context(context)
+        self.context_changed.emit(context)
+
+    def _on_entity_activated(self, entity_type, entity_id, entity_name):
+        """
+        Slot called when an entity is selected via one of the search completers.
+        """
+        publisher = sgtk.platform.current_bundle()
```
josh-t: same comment about `self._bundle`.

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+        entity_id = "%s_%s" % (
+            context.entity["type"],
+            context.entity["id"],
+        )
+
+        # if we've queried tasks for this entity before, just return those
+        if entity_id in self._related_tasks_cache:
+            return self._related_tasks_cache[entity_id]
+
+        publisher = sgtk.platform.current_bundle()
```
josh-t: same comment about `self._bundle`.

python/context_selector/context_widget.py
```diff
@@ -0,0 +1,956 @@
...
+    else:
+        return ""
+
+
+def _query_my_tasks():
+    """
+    Called via bg task to query SG for tasks assigned to the current user.
+    """
+
+    publisher = sgtk.platform.current_bundle()
```
josh-t: we should rename this

python/context_selector/resources/context_editor_widget.ui
```diff
@@ -0,0 +1,368 @@
...
+             </size>
+            </property>
+            <property name="maximumSize">
+             <size>
+              <width>16777215</width>
+              <height>32</height>
+             </size>
+            </property>
+            <property name="toolTip">
+             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;The task that the selected item will be associated with in Shotgun after publishing. It is recommended to always fill out the Task field when publishing. The menu button to the right will provide suggestions for Tasks to publish to including the Tasks assigned to you, recently used Tasks, and Tasks related to the entity Link populated in the field below. &lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
josh-t: We should probably reword this since it's not specific to publishing anymore. Perhaps expose this via the widget API so the publisher can define this.

python/context_selector/resources/context_editor_widget.ui
```diff
@@ -0,0 +1,368 @@
...
+               </size>
+              </property>
+              <property name="maximumSize">
+               <size>
+                <width>16777215</width>
+                <height>32</height>
+               </size>
+              </property>
+              <property name="toolTip">
+               <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Type here to search for a Task to associate with the selected item. It is recommended that you always populate the Task field when publishing. Once a Task is found, the Link field will be populated below with the entity to which the Task belongs.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
josh-t: Same comment here about not being publish specific.

python/context_selector/resources/context_editor_widget.ui
```diff
@@ -0,0 +1,368 @@
...
+             </size>
+            </property>
+            <property name="maximumSize">
+             <size>
+              <width>16777215</width>
+              <height>32</height>
+             </size>
+            </property>
+            <property name="toolTip">
+             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;The entity that the selected item will be associated with in Shotgun after publishing. By selecting a Task in the field above, the Link will automatically be populated. It is recommended that you always populate the Task field when publishing. The Task menu above will display any tasks associated with the entity populated in this field.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
josh-t: Same comment.

python/context_selector/resources/context_editor_widget.ui
```diff
@@ -0,0 +1,368 @@
...
+             </size>
+            </property>
+            <property name="maximumSize">
+             <size>
+              <width>16777215</width>
+              <height>32</height>
+             </size>
+            </property>
+            <property name="toolTip">
+             <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Type here to search for an entity to associate with the selected item. It is recommended that you also always populate the Task field above when publishing. Once you find the entity in this field, check the Task menu above for related tasks to associate with the item.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
josh-t: Same comment.

josh-t:
Oh! I'm also wondering if we need to pull over come of the context widget style sheet settings from the publisher. It's doing some pretty non-standard stuff with `QLineEdit` + `QToolButton` to mimic a `QComboBox` IIRC. I think it might look weird without the qss we put together for the publisher. One thing we should probably do is throw together a quick demo in the demo app and see how it looks/feels. Do we have precedent for qss in qtwidgets?

python/context_selector/context_widget.py
manneohrstrom: fixed

python/context_selector/context_widget.py
manneohrstrom: done

python/context_selector/context_widget.py
manneohrstrom: all renamed to `bundle`

python/context_selector/context_widget.py
manneohrstrom: I think it's fine to keep it for now. i agree with the comment :)

python/context_selector/context_widget.py
manneohrstrom: all renamed to bundle

python/context_selector/context_widget.py
manneohrstrom: all renamed to bundle

python/context_selector/context_widget.py
manneohrstrom: all renamed to bundle

python/context_selector/context_widget.py
manneohrstrom: all renamed to bundle

python/context_selector/resources/context_editor_widget.ui
manneohrstrom: done and added `set_task_tooltip` and `set_link_tooltip` methods.

manneohrstrom:
Ready for another round of CR @josh-t !

python/context_selector/context_widget.py
manneohrstrom: new methods added!

python/context_selector/context_widget.py
```diff
@@ -303,28 +303,96 @@ def set_up(self, task_manager):
...
+        Restrict which entity types should show up in the list of matches.
+
+        :param list entity_types: List of entity types
+        """
+        logger.debug(
+            "Restricting auto completer to show the following types: %s" % entity_types
+        )
+        # construct a dictionary that the search widget expects for
+        # filtering. This is a dictionary with the entity types as keys and
+        # values a list of search filters. We don't have any filters, so we
```
josh-t: worth allowing the calling code to supply filters here?

python/context_selector/context_widget.py
```diff
@@ -939,26 +1007,28 @@ def _query_my_tasks():
...
         filters,
         fields=task_fields,
         order=order
     )
 
 
-def _query_publishedfile_entity_schema():
-    """
-    Called as bg task to query SG for the field schema for PublishedFile.entity
+def _query_publishedfile_entity_schema(entity_type, field_name):
```
josh-t: suggest name change. perhaps `_query_linked_entity_types`?

manneohrstrom: ahh yes good point! :)

python/context_selector/context_widget.py
manneohrstrom: that syntax is a bit iffy IIRC - i'd rather leave that out for the time being - we can always add it in later! Good point though :)

PR #84, opened 2018-01-08
For #38483, edge case fix.

sgsteph:
* Fixed recursive searching for the case where an item reports having children, but hasn't loaded them yet, so the `rowCount` is 0 in the next recursion by ensuring the parent acceptance is never lost.
* Made the logic easier to understand.

### DIFF
--- a/python/models/hierarchical_filtering_proxy_model.py
+++ b/python/models/hierarchical_filtering_proxy_model.py
@@ -358,32 +358,34 @@ class HierarchicalFilteringProxyModel(QtGui.QSortFilterProxyModel):
         model = idx.model()
 
         # check to see if any children of this item are known to have been accepted:
-        child_accepted = self._child_accepted_cache.get(idx)
-        if child_accepted != None:
+        cached_value = self._child_accepted_cache.get(idx)
+        if cached_value is not None:
             # we already computed this so just return the result
-            return child_accepted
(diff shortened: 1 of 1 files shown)


sgsteph:
I do agree that the code is not simple and clear, this is why I am extremely wary of changing anything without revisiting the whole logic... 
There are two caches, the `_accepted_cache` and the `_child_accepted_cache`, which can be problematic for huge models.
I'm not super familiar with `QtGui.QSortFilterProxyModel`, but I wouldn't expect `filterAcceptsRow` to be called on an item if its parent was not accepted, so we shouldn't have to traverse the tree up when considering an item. Now, at any given level, we want to check descendants and if at least one of them is accepted, accept the current item. The problem is that by accepting this item, `filterAcceptsRow` will be called on all descendants, calling the acceptance method again on them, hence the need of a cache. I could be wrong, but I _think_ we should only need a _rejected_ cache. At any given intermediate level `filterAcceptsRow` should check if the index in the _rejected_ cache and reject it, blindly accept it otherwise, something like that. The _rejected_ cache would only need to contain the top most rejected indexes: if at any given level all direct children are rejected, just keep trace of the pruned branch and discard the cache for children as it will never be hit.
I guess what I'm saying is that we should revisit this code rather than painfully trying to make it clearer.

PR #85, opened 2018-01-10
38483 revisited filter proxy model

sgsteph:
Following the previous PR for the `HierarchicalFilteringProxyModel`, I revisited the implementation to make it simpler and clearer.

I now use a single rejection cache, instead of two acceptance caches. This cache can be sparse: to prune a branch, only the first item on this branch needs to be rejected. Items which are under an accepted node and didn't contribute to this node acceptance don't need to have cache entries.

Row deletions is now supported, and the proxy is always in sync with the cache, thanks a to a deferred refresh. Because just re-applying the existing cache to the proxy model can be slow, there is some simple code to ensure that a single refresh is posted at any time.

In terms of performances the new implementation seems very similar to the previous one: not better, but not worse. So the primary goal of these changes is to make the code easier to maintain.


python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -23,12 +26,12 @@ class HierarchicalFilteringProxyModel(QtGui.QSortFilterProxyModel):
     """
     
     
-    class _IndexAcceptedCache(object):
+    class _IndexRejectedCache(object):
         """
-        Cached 'accepted' values for indexes.  Uses a dictionary that maps a key to a tuple 
-        containing a QPersistentModelIndex for the index and its accepted value.
+        Cached 'rejected' values for indexes.  Uses a dictionary that maps a key to a tuple
```
manneohrstrom: would be great to know what 'rejected' means? Perhaps a practical example of where this is being used could help.

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -267,56 +277,50 @@ def filterAcceptsRow(self, src_row, src_parent_idx):
...
-            # even though the parent wasn't accepted, it may still be needed if one or more
-            # children/grandchildren/etc. are accepted:
-            return self._is_child_accepted_r(src_idx, parent_accepted)
+        # Check if we are dealing with top nodes
+        if not src_parent_idx.parent() or not src_parent_idx.parent().isValid():
+            self._update_rejection_cache(src_idx)
+            # We can now just return the cached value which was may be modified
+            # when traversing the tree.
+            cached_value = self._rejected_cache.get(src_idx)
+            return not cached_value
```
manneohrstrom: Minor: the `not cached_value` is confusing especially considering the three return values `True` `False` or `None`.

Could we do something like

```
_IndexRejectedCache.is_accpted(index) # raises KeyError or something in cache if value not present?
_IndexRejectedCache.is_rejected(index) # raises KeyError or something in cache if value not present?
_IndexRejectedCache.is_cached(index)
```

Perhaps that leads to clearer code?

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -329,104 +333,122 @@ def setSourceModel(self, model):
...
+        to force a refresh.
 
-        # check to see if any children of this item are known to have been accepted:
-        cached_value = self._child_accepted_cache.get(idx)
+        :param from_index: A :class:`QModelIndex` instance.
+        """
+        if not from_index or not from_index.isValid():
+            return
+        # Do we have a value from a previous iteration?
+        cached_value = self._rejected_cache.get(from_index)
```
manneohrstrom: With the suggestion above:
```
if self._rejected_cache.is_cached(from_index):
   return self._rejected_cache.is_accepted(from_index)
```

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -471,8 +532,41 @@ def _on_source_model_rows_inserted(self, parent_idx, start, end):
             # invalid input parameters so ignore!
             return
 
-        # dirty specific rows in the caches:
-        self._dirty_accepted_rows(parent_idx, start, end)
+        # Get the current value for the parent
+        cached_value = self._rejected_cache.get(parent_idx)
+        # Clear the cache value for the parent.
+        self._rejected_cache.remove(parent_idx)
```
manneohrstrom: what about siblings? are they removed too? Maybe worth clarifying.

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -329,104 +333,122 @@ def setSourceModel(self, model):
...
+                    child_index.parent(),
+                    False
+                )
+                rejected = not accepted
+                self._rejected_cache.add(child_index, rejected)
+            if rejected:
+                # We need to check children, if any, add them in the
+                # descendant list.
+                for row_i in range(child_index.model().rowCount(child_index)):
+                    descendant_indexes.append(child_index.child(row_i, 0))
```
manneohrstrom: why not make the method recursive? i personally find this pretty hard to read. Is it an optimization of sorts?

sgsteph: it is not recursive because it does not need to be: we don't need to get back results from children recursions. The cache is updated bottom to top around line 409, and the update can go higher in hierarchy than the starting point. Usually iterative approaches are more efficient than recursive approaches. When comparing the previous approach with this one, I didn't spot any noticeable differences for a Sequence/Shot/Step/Task tree, ~ 5000 Tasks, so I guess it could be turned into a recursive approach if you feel strongly about it?

python/models/hierarchical_filtering_proxy_model.py
sgsteph: will add a comment, siblings are not affected.

python/models/hierarchical_filtering_proxy_model.py
sgsteph: Will clarify.

python/models/hierarchical_filtering_proxy_model.py
sgsteph: I personally find  `self._rejected_cache.is_accepted` confusing. I agree that `return not cached_value` is not great either... There is a notion of being _implicitly_ accepted for some items which are not in the cache. Let me see if I can make that clearer.

python/models/hierarchical_filtering_proxy_model.py
manneohrstrom: not feeling strongly! Whatever you feel is best!

jfboismenu: Steph, did you test with something a bigger than 5000 tasks and see how your code scales with bigger payloads? I remember originally the time to build the tree used to be exponential in wf2.

jfboismenu:
I'll wait for Stéphane's update to the code if you all don't mind before commenting.

sgsteph:
Pushed some improvements and a bug fix. Hopefully things are clearer now, let me know!

PR #88, opened 2018-03-27
For #46757, gentler error when `tank_name` is not set.

jfboismenu:
Refactored overlay widget into a QLabel so we can insert hyperlinks.

See [this PR](https://github.com/shotgunsoftware/tk-desktop/pull/96) to see it in action.

Here's a screenshot of the updated doc.
screen shot 2018-03-27 at 08 50 13

### DIFF
--- a/python/overlay_widget/shotgun_overlay_widget.py
+++ b/python/overlay_widget/shotgun_overlay_widget.py
@@ -20,6 +20,13 @@ class ShotgunOverlayWidget(QtGui.QLabel):
     Overlay widget that can be placed on top over any QT widget.
     Once you have placed the overlay widget, you can use it to
     display information, errors, a spinner etc.
+
+    The :meth:`show_message` and :meth:`show_error_message` accept
+    both regular text and HTML to format the error message.
+
+    Constants ``INFO_COLOR`` and ``ERROR_COLOR`` are provided on the class
+    as shorthand for the colors employed by the :meth:`show_message` and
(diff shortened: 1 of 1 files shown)


python/overlay_widget/shotgun_overlay_widget.py
```diff
@@ -45,16 +55,18 @@ def __init__(self, parent):
...
+        self.setWordWrap(True)
 
-        # turn off the widget
-        self.setVisible(False)
-        self._mode = self.MODE_OFF
+        # Allow to open hyperlinks
+        self.setOpenExternalLinks(True)
+
+        # Dark gray background.
+        self.setStyleSheet("background-color: #1B1B1B")
```
josh-t: Not to add too much more work, but I think the convention would be to add a `style.qss` file into this widget's folder and apply it when the widget is created. This avoids hardcoding the style in the code.

josh-t: You might be able to set the label text with div or span tags and use class selectors in the qss file to handle the error vs. info colors. i.e.: `label.setText("<div className='error'>Error message</div>")` ... then, in the qss file:

```
.error {
    color: red;
}
```

I haven't tried this before actually so not sure if it will work.

jfboismenu: I've looked at the Qt doc and it doesn't seem like you can specify contants in that file. You can only have selectors.

PR #90, opened 2018-09-14
SG-6409: Doesn't hide/show the activity stream during a clear operation.

thebeeland:
We've been experiencing stability issues with the activity stream widget in PySide2 environments that's tied to this hide/show routine that's part of the widget's clear method. The change here is to only do that show/hide when Qt4 is in use, and to skip it otherwise.

One secondary change is that the activity stream widget now registers the task manager it's given with shotgun_globals. Since the widget makes use of shotgun_globals extensively, it should ensure that it has access to a background task manager.


thebeeland:
The pattern does exist in other places, but for whatever reason it's only during the clear routine that it's causing stability problems.  So it's not the hide/do something/show that's the problem, it's the specific combination of hide/show wrapping portions of this clear process.

PR #96, opened 2019-09-27
Added HumanUser Group to the filters list for My Tasks

rob-aitchison:
In tk-multi-publish2, the My Tasks context selector would not show Tasks assigned to a Group that the User belonged to. I modified the filter so that it would.


pscadding:
It looks good to me!

PR #99, opened 2019-10-30
SG-13266: port to Python 3

jfboismenu:
Ported the framework to Python 3. Note that the updates to the `build_resources.sh` script and the resources are on [another branch](https://github.com/shotgunsoftware/tk-framework-qtwidgets/compare/SG-13266_clean_review?expand=1), otherwise there would be another 100+ files in the PR.

I added a test that loads the framework and instantiates all the widgets one by one. I actually found some bugs with it, so it was useful.

Then I started working on the multi demo app, which helped me find more issues, which I've fixed in this branch.


python/activity_stream/data_manager.py
```diff
@@ -722,37 +726,39 @@ def __db_insert_activity_updates(
                         INSERT OR REPLACE INTO activity(activity_id, payload, created_at)
                         SELECT ?, ?, datetime('now')
                     """
+                    params = (activity_id, blob)
```
jfboismenu: This was actually a bug in the current implementation! We passed too many bindings in.

python/models/hierarchical_filtering_proxy_model.py
```diff
@@ -141,7 +142,7 @@ def minimize(self):
                 return
 
             self._cache = dict(
-                [(k, v) for k, v in self._cache.iteritems() if v[0].isValid()]
+                [(k, v) for k, v in six.iteritems(self._cache) if v[0].isValid()]
```
jfboismenu: Filtering model need to be super-fast, so I'm keeping the use of iterators here.

tests/fixtures/config/env/test.yml
```diff
@@ -14,7 +14,7 @@ engines:
       type: path
       path: $SHOTGUN_TEST_ENGINE
 frameworks:
-  tk-framework-qtwidgets:
+  tk-framework-qtwidgets_v2.x.x:
```
jfboismenu: This is from the test that imports everything I mentionned in the PR description.

tests/test_framework.py
```diff
@@ -8,22 +8,147 @@
...
+                        params["bg_task_manager"] = bg_task_manager
+                    elif arg == "fields_manager":
+                        params["fields_manager"] = field_manager.ShotgunFieldManager(
+                            parent=parent, bg_task_manager=bg_task_manager
+                        )
+                    elif arg == "sg_entity_type":
+                        params["sg_entity_type"] = "Asset"
+
+                # Add the widget and show it, don't worry about the rest.
+                # It. Just. Works.
```
willis102: 😂 ❤️

PR #107, opened 2020-08-20
Bug Fix: GroupedListView bad index on _get_item_rect

staceyoue:
Handle the case when `_get_item_rect` is passed an invalid index or it is the root index -- this will create an empty `rows` list which will error on `rows[-1]`


staceyoue:
> Could you elaborate on what exactly is happening? It seems like at least one caller of _get_item_rect() checks if the index is valid before calling the method. In which situation are we calling the method with an invalid index?

I'm working to resolve style conflicts with Sketchbook, so I'm calling `QWidget.setStyle()` on a `BrowserForm` object, which then triggers a call to `GroupListView::scrollTo()`, and an invalid index sometimes is passed.

PR #109, opened 2020-11-04
Expose additional components in search_completer

Ahuge:
This exposes some of the hidden modules that were previously impossible to import using `import_framework`

This allows clients to take each part of this system and potentially subclass them and swap different components in and out.

Internally in one of my own apps, I need to create my own subclass of `global_search_result_delegate.GlobalSearchResultDelegate` because the GlobalSearchCompleter when it runs the `shotgun_api3.Shotgun.text_search` it doesn't return the links result that my app needs.

Because of the way GlobalSearchCompleter works, I need to do a bunch of subclassing, and the `global_search_result_delegate` submodule was not exported by the `__init __` file.

### DIFF
--- a/python/search_completer/__init__.py
+++ b/python/search_completer/__init__.py
@@ -9,4 +9,10 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 from .global_search_completer import GlobalSearchCompleter
+from .global_search_result_delegate import GlobalSearchResultDelegate
 from .hierarchical_search_completer import HierarchicalSearchCompleter
+from .hierarchical_search_result_delegate import HierarchicalSearchResultDelegate
+from .search_completer import SearchCompleter
+from .search_result_delegate import SearchResultDelegate
+from .search_result_widget import SearchResultWidget
+from .utils import create_rectangular_thumbnail, CompleterPixmaps


jfboismenu:
Hey @Ahuge ,

We're sorry, but in order for us to expose that part of the API, we would have to fully document it and decide what we actually want to allow clients to extend. We have no plan to do this at the moment.

Could you share a bit more with us what you did, maybe screenshots or code snippets, so we have a better understanding of your custom needs. That would help us in the future if we decided to allow clients to expand on this.

I'm sorry we can't be of any more help at the moment.

Ahuge:
Hey JF,

So what I am trying to do is expose the linked entity

My existing code, using `shotgun_fields.entity_widget.EntityEditorWidget` when running searching against a "Version.entity" field, allows me to search for Shot, Sequence, Asset, etc. However when I do that, because the `shotgun_api3.Shotgun.text_search` call doesn't return any linked entities, I get the following. Which is often fine, but in the case below, the project has several shots with the same name.
image

After my modification that uses these unexposed classes, I get the following, which shows the linked Sequence off of a Shot
image (3)


I started patching the `GlobalSearchResultDelegate` so that if there were no links returned from the Shotgun.text_search call, it would attempt to populated the returned entity based on "interesting link types" which I have hardcoded to be `("Shot", "Sequence", "Episode", "Project", "Asset")` basically.
Now that is quite specific to us, and I don't have a more generic way to implement that into the `GlobalSearchResultDelegate`, which is fine.

My client code implements that basically reimplements the `_render_result` protected method and I am ok with taking responsibility for utilizing your non-public api in that case.

I update reimplement the GlobalSearchCompleter with my custom delegate and then set a custom completer in my custom `entity_widget.EntityEditorWidget`.
The problem is the GlobalSearchCompleter and GlobalSearchDelegate (among other classes), aren't exposed so I cannot reimplement them, or even use those components in totally separate use cases.


I guess a different way to approach this would be if the `shotgun_api3.Shotgun.text_search` call returned more appropriate `links` data. I am not 100% sure what the logic is there, but it's all server side, so I decided this was less invasive for the broader Shotgun ecosystem.

Perhaps there is something simple and obvious that I should have done instead?

Ahuge:
Also, I totally understand you not wanting to add more burden.  


What I am going to do is basically vendor the required modules and separate out all of the vanila tk-framework-qtwidgets code. I will also keep my patched implementations separate from that in case this ever is able to be exposed I can transition back to the qt-framework-qtwidgets code from the framework.

Thanks
-Alex

PR #110, opened 2020-11-11
SG-20166: global search unicode filter fix

jfboismenu:
This fixes an issue where global search results with a unicode character in them would completely fail to render. Now the result will be visible and selectable. However, due to the broader issue of incorrect unicode rendering in Toolkit in Python 2, the entry search results is going to keep garbled characters instead of the expected character, as can be seen in the screenshot below.

image

Note, the hierarchy search completer does not exhibit this bug and can properly match entries with unicode characters, so no fix was necessary.

Also, the rendering of unicode chars does not seem to be an issue in Python 3. 🙌 

image

### DIFF
--- a/python/search_completer/search_result_delegate.py
+++ b/python/search_completer/search_result_delegate.py
@@ -9,6 +9,7 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sgtk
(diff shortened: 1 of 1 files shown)


pscadding (approved):
Yeah nice, looks good!

PR #125, opened 2022-07-01
[SHOT-4077] Removed :inherited-members: from all .rst files

rob-aitchison:


barbara-darkshot:
hey @rob-aitchison , do we know what has introduced the changes? hopefully, if removing the inherited members makes the trick it's nice 👍 just curious why it was suddenly broken when the files haven't been changed since a very long time

rob-aitchison:
> hey @rob-aitchison , do we know what has introduced the changes? hopefully, if removing the inherited members makes the trick it's nice 👍 just curious why it was suddenly broken when the files haven't been changed since a very long time

Hey @barbara-darkshot , I did not research _why_ yet but it would be good to track it down to know why, I agree. Maybe the sphinx changelog?

barbara-darkshot (approved):
LGTM 👍

PR #128, opened 2022-08-03
Bug Fix for Filter Menu

staceyoue:
* Do not assume all data given is valid, check that field id keys exist first


000paradox000 (approved):
LGTM

PR #129, opened 2022-08-16
Bug fix: Grouped item view selection

staceyoue:
* Need to continue after selecting the group to ensure the y offset is updated to avoid selecting indexes that are not in the selection intersection


000paradox000 (approved):
LGTM

PR #130, opened 2022-08-29
SG-27819: Fix duplicate filter name

staceyoue:
* Filter names are built based on the entity display and field display names, but there are some cases where the name will be duplicated if the field display name already includes the entity name. In this case (where the entity name is already included in the field display), we will append the field name (not display name) in brackets


NorberMV (approved):
LGTM!

PR #133, opened 2022-12-12
SG-29407: Improve delegate handing of mouse mouse events.

staceyoue:
* Do not emit data changed signal on mouse move event, this is not necessary and causes the UI responsiveness to be slow (due to extra paint events being triggered)
* We cannot use the `roles` parameter to specify what data has changed when emitting the `dataChanged` signal since tk-core patches this method and ignores this parameter
* Fix method call


NorberMV (approved):
LGTM !

PR #134, opened 2023-01-06
Add new decorators module

staceyoue:
* New module to access common decorator functions across Toolkit Apps
* Added decorator function to show busy/wait cursor during the execution of a function
* Show busy cursor while filter menu refreshes

### DIFF
--- a/python/decorators/decorators.py
+++ b/python/decorators/decorators.py
@@ -1,0 +1,34 @@
+# Copyright (c) 2022 Autodesk, Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the ShotGrid Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the ShotGrid Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Autodesk, Inc.
+
+from functools import wraps
+
+import sgtk
+from sgtk.platform.qt import QtGui, QtCore
+
+
+def wait_cursor(func):
+    """
+    Decorator function that overrides the Qt cursor to show the waiting cursor while a function executes.
+
+    :param func: The function to execute
+    :type func: function
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        QtGui.QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
+        try:
+            response = func(*args, **kwargs)
+        finally:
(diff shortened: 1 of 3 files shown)


NorberMV (approved):
Nice feature, LGTM !

python/decorators/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+
+from functools import wraps
+
+import sgtk
+from sgtk.platform.qt import QtGui, QtCore
+
+
+def wait_cursor(func):
+    """
+    Decorator function that overrides the Qt cursor to show the waiting cursor while a function executes.
```
julien-lang, line 19: Could you split the line at 80 (or 120 I can't remember) chars like the rest of the code, please?

python/decorators/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# This work is provided "AS IS" and subject to the ShotGrid Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the ShotGrid Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Autodesk, Inc.
+
+from functools import wraps
+
+import sgtk
```
julien-lang, line 13: Is this import required?

python/decorators/decorators.py
```diff
@@ -0,0 +1,34 @@
...
+
+    :param func: The function to execute
+    :type func: function
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        QtGui.QApplication.setOverrideCursor(QtCore.Qt.WaitCursor)
+        try:
+            response = func(*args, **kwargs)
```
julien-lang, line 29: The intermediary response variable is not used. You could just return here. The finally would run anyway.

julien-lang:
Just to be sure: you don't use this new feature here right? You will use it in a different PR/repo?

julien-lang (changes requested):
I left a few comments to address.

python/decorators/decorators.py
staceyoue: No it is not, removed.

staceyoue:
> Just to be sure: you don't use this new feature here right? You will use it in a different PR/repo?

Yes, it is used in this PR: https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/136

staceyoue:
> I left a few comments to address.

Thanks for the review, pushed updates for your comments.

julien-lang (approved):
LGTM!

PR #135, opened 2023-01-16
SearchWidget emit search changed signal on editing finished

staceyoue:
* Provide the option to listen for editing finished signal for whole text changes, instead of each new key stroke
* Update the `search_changed` signal to emit when the line edit sends the `editingFinished` signal

This change may help address this issue: https://community.shotgridsoftware.com/t/removing-the-search-widget-in-tk-multi-workfiles2/16478

### DIFF
--- a/python/search_widget/search_widget.py
+++ b/python/search_widget/search_widget.py
@@ -70,6 +70,7 @@ class SearchWidget(QtGui.QWidget):
 
         # hook up the signals:
         self._ui.search_edit.textEdited.connect(self._on_text_edited)
+        self._ui.search_edit.editingFinished.connect(self._on_editing_finished)
         self._ui.search_edit.returnPressed.connect(self._on_return_pressed)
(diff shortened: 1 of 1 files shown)


NorberMV (commented):
The new slots looks great! just one minor question here, so will this _on_text_edited() slot will still have any relevance for this widget?, or do we want to keep using the "search_changed" signal emited from "_on_editing_finished()" instead when using the search_widget ?

staceyoue:
> The new slots looks great! just one minor question here, so will this _on_text_edited() slot will still have any relevance for this widget?, or do we want to keep using the "search_changed" signal emited from "_on_editing_finished()" instead when using the search_widget ?

The `search_changed` signal is probably better to use in most cases, but I don't see any harm in keeping the `_on_text_edited` slot that triggers the `search_edited` signal. It gives the option to listen for keystroke changes if needed (instead of only when the text is done being edited).

NorberMV (approved):
LGTM !

PR #136, opened 2023-01-17
Filtering Improvements

staceyoue:
Improvements to filtering:
- Save and restore filter menu state
- Add option to determine whether or not to refresh on filter menu show
- Show loading spinner icon on filter menu button while refreshing
- Improve filter definition build time and add method to update definition instead of full rebuild
- Improve filter menu logic for faster build/update time
- Trigger search text filter change only once editing finished (instead of each key stroke)

Includes changes from other PRS:
- https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/135
- https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/134
- https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/133


rob-aitchison (approved):
Looks great to me, thank you!

NorberMV (commented):
Everything looks good from my perspective.

python/delegates/view_item_delegate.py
```diff
@@ -1158,13 +1158,6 @@ def editorEvent(self, event, model, option, index):
                 else:
                     widget.unsetCursor()
 
-            # Emit a data changed signal for the index to paint according to the mouse move
-            # event (e.g. hover selection).
-            # TODO: leverage the 'roles' parameter to emit only the relevant data roles
-            # to consider, for a more efficient update. Currently there are some issues using
-            # this parameter due to mismatch method signatures between versions.
-            model.dataChanged.emit(index, index)
```
NorberMV, line 1166: I wonder if we shouldn't continue issuing dataChanged() here after setting the role data here for the index.model().setData(index, new_check_state, check_state_role) element ?

python/filtering/filter_item_widget.py
```diff
@@ -356,23 +318,17 @@ def value(self, value):
         """
 
         if not isinstance(value, six.string_types):
-            assert False, "Attempt to set non-string data to QLineEdit"
```
NorberMV, line 359: Not sure why this check is being removed here, are we expecting other types of searching values ?

python/delegates/view_item_delegate.py
staceyoue: This update is actually part of the PR (that has been merged) https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/133/files

In short, this data change signal is not necessary here and was triggering more paint events than needed, which negatively affects performance

python/filtering/filter_item_widget.py
staceyoue: I originally removed the assert to just ignore it because we were getting boolean values passed in some times when check states are changed; however, you're right in asking if we should ever expect other types being passed, because we shouldn't in the first place.

I added back the assert, and then removed the signal/slot connection that was incorrectly trying to set the text filter item widgets value to the action's check state. While I was making this change, I also improved how the filter item value is set and to avoid triggering menu refresh if the value being set is the same as the current value.

NorberMV (approved):
This looks great !
LGTM!

PR #137, opened 2023-02-15
SG-21292 Ensure the user has the permission to see the note's author

dsauve-adsk:
When a user does not have access to scripts, the activity stream contains references to note that this user does not have access to and it make qt-widget to crash. So I added a test to be sure that the current_user can access the note's Author to be sure there is no issue.

### DIFF
--- a/python/activity_stream/activity_stream.py
+++ b/python/activity_stream/activity_stream.py
@@ -873,26 +873,31 @@ class ActivityStreamWidget(QtGui.QWidget):
                 widget.interactive = self.version_items_playable
 
             elif data["primary_entity"]["type"] == "Note":
-                # new note
-                widget = NoteWidget(data["primary_entity"]["id"], self)
-                widget.show_note_links = self.show_note_links
-                widget.attachments_filter = self.attachments_filter
+                # Ensure that the user has permission to see script user
+                # (if not, data["primary_entity"]["user"] will be None)
+                if data["primary_entity"]["user"]:
+                    # new note
+                    widget = NoteWidget(data["primary_entity"]["id"], self)
+                    widget.show_note_links = self.show_note_links
+                    widget.attachments_filter = self.attachments_filter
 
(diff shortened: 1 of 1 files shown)


python/activity_stream/activity_stream.py
```diff
@@ -873,26 +873,28 @@ def _create_activity_widget(self, activity_id):
...
-                    for w in self._activity_stream_data_widgets.values():
-                        if isinstance(w, NoteWidget):
-                            if w.selected:
-                                w.set_selected(False)
-                                self._note_selected_changed(False, w.note_id)
-
-                    widget.set_selected(True)
-                    self._note_selected_changed(True, widget.note_id)
-                    self._select_on_arrival = dict()
+                # Ensure that the user have permission to see script user (if not, data["primary_entity"]["user"] will be None)
```
julien-lang, line 876: have -> has

Also can you split the comment line to 80 characters like the other ones?

python/activity_stream/activity_stream.py
```diff
@@ -873,26 +873,28 @@ def _create_activity_widget(self, activity_id):
...
+                if data["primary_entity"]["user"] != None:
```
julien-lang, line 877: You can drop `!= None`. It provides the same test I believe.

python/activity_stream/activity_stream.py
dsauve-adsk: done

python/activity_stream/activity_stream.py
dsauve-adsk: done

julien-lang (approved):

julien-lang (changes requested):
Please check the test result. There seems to be a code style issue.

dsauve-adsk:
> 

All god now

julien-lang (approved):

NorberMV (approved):
LGTM!

PR #138, opened 2023-06-14
SG-30534 and SHOT-4264: Fix thumbnail view for PySide6

staceyoue:
* Do not use `isinstance` method for Qt classes, it is not guaranteed to work due to how Toolkit supports multiple versions of Qt (some Qt class may be patched, and are not the original PySide class)

### DIFF
--- a/python/delegates/view_item_delegate.py
+++ b/python/delegates/view_item_delegate.py
@@ -1603,116 +1603,116 @@ class ViewItemDelegate(QtGui.QStyledItemDelegate):
         if not badge_data:
             return
 
-        if isinstance(badge_data, QtGui.QPixmap):
-            # Default to bottom right if only a pixmap given.
+        if not isinstance(badge_data, dict):
+            # Default to bottom right if only a pixmap given for badge data.
             badge_data = {"bottom-right": badge_data}
 
-        if isinstance(badge_data, dict):
-            for position, pixmap_data in badge_data.items():
-                if not pixmap_data:
-                    continue
+        for position, pixmap_data in badge_data.items():
+            if not pixmap_data:
+                continue
 
-                # The pixmap to display for the icon badge
-                pixmap = None
-                # Inset will display the badge inside the bounding rect, otherwise the badge
-                # will be display centered over the corner corresponding to the badge position.
-                inset = True
-                if isinstance(pixmap_data, dict):
-                    pixmap = pixmap_data.get("pixmap", None)
-                    inset = pixmap_data.get("inset", True)
+            # The pixmap to display for the icon badge
+            pixmap = None
+            # Inset will display the badge inside the bounding rect, otherwise the badge
+            # will be display centered over the corner corresponding to the badge position.
+            inset = True
+            if isinstance(pixmap_data, dict):
+                pixmap = pixmap_data.get("pixmap", None)
+                inset = pixmap_data.get("inset", True)
+            else:
+                pixmap = pixmap_data
+
+            if hasattr(pixmap, "pixmap"):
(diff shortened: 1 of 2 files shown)


rob-aitchison (approved):
Excellent, thank you!

python/delegates/view_item_delegate.py
```diff
@@ -1603,116 +1603,116 @@ def _draw_icon_badges(self, painter, option, bounding_rect, index):
...
+            if hasattr(pixmap, "pixmap"):
+                pixmap = self._convert_icon_to_pixmap(pixmap)
+            elif isinstance(pixmap, six.string_types):
+                pixmap = QtGui.QPixmap(pixmap)
 
-                if not pixmap or not isinstance(pixmap, QtGui.QPixmap):
-                    # Skip, invalid pixmap data.
-                    continue
+            # if not pixmap or not isinstance(pixmap, QtGui.QPixmap):
+            if not pixmap:
```
julien-lang, line 1631: Hum, looks like something is not finished here with the commented line above.

python/delegates/view_item_delegate.py
```diff
@@ -3554,15 +3551,10 @@ def set_icon(self, icon):
 
         if not icon:
             self.icon = None
-
-        elif isinstance(icon, QtGui.QIcon):
```
julien-lang, line 3558: Why not replacing to `if hasattr(icon, "pixmap"):` like you did before instead of a "else" block

julien-lang (changes requested):
Looks good but I think there is one change to make (cf comments).

Also, is it not possible to combine multiple `isinstance` call (one for qt4-5 and one for qt6) instead of using the `hasattr` heuristic?

staceyoue:
> Also, is it not possible to combine multiple `isinstance` call (one for qt4-5 and one for qt6) instead of using the `hasattr` heuristic?

@julien-lang 

So the issue is that we patch QPixmap class constructor for qt6, which means if a QPixmap is created using `__init__` then it will have the patched QPixmap class. However, if we get a pixmap object from say, `QIcon.pixmap()`, the type of this object is the original PySide6 QPixmap class. This issue is not limited to QPixmap, it will happen for any Qt class that is patched.

This could be resolved by checking multiple classes in the isinstance call, though it gets ugly with having to try/except import PySide to get the correct PySide version. An alternative could be to add a method to the `sgtk.platform.qt` module to check `isinstance`, this way we already have the correct PySide version to use. This though would require a lot more work to the QtImporter class.

python/delegates/view_item_delegate.py
staceyoue: I didn't use `hasattr(icon, "pixmap")` here since we don't actually use the `QIcon.pixmap` method.

This way puts the onus on the caller to pass the correct object for the icon, instead of checking if it is a valid icon and if not, setting the icon to None. Arguably this way might be better since if the icon is invalid, you'll get an error message down the line instead of silently just setting the icon to None.

python/delegates/view_item_delegate.py
staceyoue: Oops, forgot to remove that commented line. I am removing the sanity check that the pixmap is a valid QPixmap, similar to below about the QIcon check, it may actually be better to put the onus on the caller to pass the correct object type and if not, get an error message instead of silently passing on.

julien-lang:
@staceyoue I would prefer the `isinstance` approach but I understand it would take you a lot more work so I am not going to make it a blocker.

carlos-villavicencio-adsk (approved):
LGTM

julien-lang (approved):
I'm fine merging the PR as-is. The left over comments will be addressed when dropping QT4.

PR #139, opened 2023-07-28
SHOT-4273: Fix save and restoring filters

staceyoue:


carlos-villavicencio-adsk (approved):

staceyoue:
Found a small issue, addressed in last two commits.

PR #140, opened 2023-07-31
SHOT-4272: Filter group menu actions

staceyoue:
* Add filter group menu actions to reset and remove a filter group
* Reset will clear all filters within the filter group (this provides the functionality of the "None" published file type filter in the previous filter component that the filter menu replaced)
* Remove will reset and remove the filter widget (within the group) from the menu


python/filtering/filter_menu_group.py
```diff
@@ -267,8 +274,36 @@ def add_to_menu(
...
 
         if title:
-            title_action = menu.add_label(title)
-            self.header_action = title_action
+            # Add the filter group title menu entry (with actions)
+            reset_action = QtGui.QAction("Reset")
+            reset_action.triggered.connect(
+                lambda checked=False, m=menu: self.reset_filters(m)
+            )
+            remove_action = QtGui.QAction("Remove")
```
julien-lang, line 282: I find the meaning of reset and remove quite confusing. Is it possible to clarify with more words (reset to default, remove all) and/or add tooltips with information?

julien-lang (commented):
Looks good from what I understand.

Question: does this create a change of behavior "by default"? I mean without any user interaction or setting, when this version will be published, will the view be filtered in a different way?

python/filtering/filter_menu_group.py
```diff
@@ -9,18 +9,25 @@
...
-    """
-    Class object to manage a filter grouping within a QMenu.
-    """
+    """Class object to manage a filter grouping within a QMenu."""
 
     def __init__(self, group_id, show_limit_increment=5):
-        """
-        Constructor. Initialize the filter group's instance members.
-        """
+        """Constructor. Initialize the filter group's instance members."""
```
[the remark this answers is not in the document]
staceyoue: I think this was a preference from a previous Toolkit member. I reverted sine it was not necessary (nor did black complain)

staceyoue:
> Question: does this create a change of behavior "by default"? I mean without any user interaction or setting, when this version will be published, will the view be filtered in a different way?

No, the view will remain filtered in the same way when this change is published. It just adds menu options to change the existing filters.

image

python/filtering/filter_menu_group.py
julien-lang: OK, that makes sense. Yes, please rename to "Reset Filter" and "Remove Filter". That will be enough for me.
Thanks.

julien-lang (approved):
LGTM!

PR #146, opened 2023-09-18
Bug Fix:  Ensure context has entity name data, for the context widget

staceyoue:
* In some cases, like for custom entities, the context retrieved does not set the `name` on the entity data. This `name` field is expected, so an error occurs later when trying to display the selected entity if it has no name data


julien-lang (approved):
LGTM!

PR #148, opened 2023-11-24
SHOT-4320:  Bug fix save and restore filters

staceyoue:
* Save and restore filters bug found in tk-multi-data-validation app
* Restore filters only after the filter definition has been fully rebuilt (else restore may fail if the filter definition does not exist yet)


carlos-villavicencio-adsk (approved):
LGTM

PR #149, opened 2023-12-04
SHOT-4296: Sort filters when displayed in menu

staceyoue:
* Add `sort_value` property to FilterItemWidget class to allow widgets to be sorted by their actual value (not by display value)
* This fixes displaying numeric filter widgets, where their value is numerical but display is a string representation (this caused numeric filters to be sorted like: 1, 11, 2, 22, 3, 4 instead of 1, 2, 3, 4, 11, 22)

### DIFF
--- a/python/filtering/filter_item_widget.py
+++ b/python/filtering/filter_item_widget.py
@@ -29,7 +29,7 @@ class FilterItemWidget(QtGui.QWidget):
     # Signal emitted when the filter widget's value changed.
     value_changed = QtCore.Signal(object)
 
-    def __init__(self, filter_id, group_id, parent=None, bg_task_manager=None):
+    def __init__(self, filter_id, group_id, parent=None, bg_task_manager=None, filter_data=None):
         """
         Constructor. Set up the widget.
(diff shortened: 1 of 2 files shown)


python/filtering/filter_item_widget.py
```diff
@@ -91,19 +99,31 @@ def value(self):
...
 
+    @property
+    def sort_value(self):
+        """Get the widget's value that is sortable."""
+        # First try to extract the sort value from the raw value.
+        value = self._raw_value
+        if isinstance(value, dict):
+            value = value.get("name")
+        # Default to the display name if the raw value is not sortable.
+        if value is None:
```
julien-lang, line 110: You might prefer `if not value` which will match with both None value and empty string.

julien-lang (approved):

python/filtering/filter_item_widget.py
staceyoue: The empty string is ok here, we want to use the raw value to sort as long as it is a sortable value.

PR #150, opened 2023-12-07
Config 1.6.1 RC3 - Linux - The Loader and the Shotgun panel fail to load and the Publishing does not work

NorberMV:
This PR addresses a bug encountered during the testing of the config v1.6.1-rc3 . This bug caused a SyntaxError in older versions of Houdini (17.5.460) and Nuke (12.0v8), which both still rely on Python 2.7.

### DIFF
--- a/python/filtering/filter_menu.py
+++ b/python/filtering/filter_menu.py
@@ -1141,7 +1141,7 @@ class FilterMenu(NoCloseOnActionTriggerShotgunMenu):
         :return: The field id that the search filter item widget refers to.
         :rtype: str
         """
-        return re.sub(rf".{str(SearchFilterItemWidget)}$", "", filter_id)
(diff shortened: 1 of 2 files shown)


python/filtering/filter_menu.py
```diff
@@ -1141,7 +1141,7 @@ def _get_search_filter_field_id(filter_id):
         :return: The field id that the search filter item widget refers to.
         :rtype: str
         """
-        return re.sub(rf".{str(SearchFilterItemWidget)}$", "", filter_id)
+        return re.sub(r"{}$".format(str(SearchFilterItemWidget)), "", filter_id)
```
NorberMV, line 1144: Caught another Syntax error from this line` rf".{str(SearchFilterItemWidget)}$"`.  Everything's working fine now.

PR #154, opened 2024-01-22
Bug fix to get the correct field id from the search widget

staceyoue:
* Add missing '.' after fix for Python 2 syntax.


carlos-villavicencio-adsk (approved):
LGTM

PR #161, opened 2024-03-28
SG-34637 qt6  mac closing multi workfiles2 makes maya crash

NorberMV:
This is a workaround for the issue in Maya where it crashes when attempting to disconnect signals in the closeEvent method. It seems that the signals were already being disconnected. After handling the exception, the application's functionality remains unaffected.


carlos-villavicencio-adsk (approved):
:shipit:

PR #162, opened 2024-05-21
SG-35252 Update conditional to support PySide6 parent

carlos-villavicencio-adsk:
This is required for https://github.com/shotgunsoftware/tk-multi-demo/pull/36

- In PySide6, `value` is type `<class 'PySide6.QtGui.QPixmap'>` instead.  A tk-core patched method wraps and uses the expected "patched" class to build a pixmap object
- Update Enum usage

**TODO**

- [x] Merge and release: https://github.com/shotgunsoftware/tk-core/pull/976
- [x] Update info.yml

### DIFF
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -12,6 +12,7 @@ import os
 
 import sgtk
 from sgtk.platform.qt import QtCore, QtGui
+from sgtk.platform.util import is_qt_instance
 from tank_vendor import six
 from .shotgun_field_meta import ShotgunFieldMeta
 
@@ -305,7 +306,8 @@ class ImageWidget(QtGui.QLabel):
             In delegate mode, this value can also be an existing ``QPixmap``
             object.
         """
-        if type(value).__name__ == "QPixmap":
+        # if isinstance(value, QtGui.QPixmap) or type(value).__name__ == "QPixmap":
+        if is_qt_instance(value, QtGui.QPixmap):
             self.setPixmap(value)
         elif os.path.exists(value):
             # a local path has been set as the value.
@@ -401,7 +403,6 @@ class ImageWidget(QtGui.QLabel):
 
         Display the image in a transient, modeless QDialog.
         """
-
         # don't continue unless there's somethign to show
         if not self._pixmap:
             return
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -306,7 +306,7 @@ class ImageWidget(QtGui.QLabel):
             In delegate mode, this value can also be an existing ``QPixmap``
             object.
         """
-        if is_qt_instance(value, QtGui.QPixmap):
+        if isinstance(value, QtGui.QPixmap):
             self.setPixmap(value)
         elif os.path.exists(value):
             # a local path has been set as the value.
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -306,7 +306,6 @@ class ImageWidget(QtGui.QLabel):
             In delegate mode, this value can also be an existing ``QPixmap``
             object.
         """
-        # if isinstance(value, QtGui.QPixmap) or type(value).__name__ == "QPixmap":
         if is_qt_instance(value, QtGui.QPixmap):
             self.setPixmap(value)
         elif os.path.exists(value):
--- a/python/shotgun_fields/image_widget.py
+++ b/python/shotgun_fields/image_widget.py
@@ -305,6 +305,7 @@ class ImageWidget(QtGui.QLabel):
             In delegate mode, this value can also be an existing ``QPixmap``
             object.
         """
+
         if isinstance(value, QtGui.QPixmap):
             self.setPixmap(value)
         elif os.path.exists(value):
@@ -401,6 +402,7 @@ class ImageWidget(QtGui.QLabel):
 
         Display the image in a transient, modeless QDialog.
         """
+
         # don't continue unless there's somethign to show
         if not self._pixmap:
             return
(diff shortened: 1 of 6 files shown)


python/shotgun_fields/image_widget.py
```diff
@@ -305,8 +305,7 @@ def _display_value(self, value):
             In delegate mode, this value can also be an existing ``QPixmap``
             object.
         """
-
-        if isinstance(value, QtGui.QPixmap):
+        if type(value).__name__ == "QPixmap":
```
julien-lang, line 308: Can we use the tk-core patcher class instead?

carlos-villavicencio-adsk: In this file, `QtGui` is imported `from sgtk.platform.qt`, Can you elaborate more?

staceyoue: > In PySide6, value is type <class 'PySide6.QtGui.QPixmap'> instead.

Which qt "interface" are you using? Is this to start porting to QT6 from QT4 interface?

carlos-villavicencio-adsk: @staceyoue I think we're still using QT4 interface. This change is because there's a class mismatch when it's running on PySide6, detected by tk-multi-demo.

staceyoue: Ok yes, we are still using QT4 interface then but with PySide6 under the hood.

The reason this gives you an error is because for PySide6 with QT4 interface we patch the QPixmap class [here](https://github.com/shotgunsoftware/tk-core/blob/67cb1bccbe3b0c5743ddbeb312e42409366c3b47/python/tank/util/pyside6_patcher.py#L88-L111) but some how the tk-multi-demo is creating the QPixmap object not through this patched class, instead it is creating it through PySide6 original QPixmap class, which is why you see `<class 'PySide6.QtGui.QPixmap'>` when it should be something like `<class 'sgtk.platform.qt.QtGui.QPixmap'>`.

carlos-villavicencio-adsk: Now I used a new util method introduced on https://github.com/shotgunsoftware/tk-core/pull/975

carlos-villavicencio-adsk:
@staceyoue @julien-lang turns out the standard model's [icon](https://doc.qt.io/qtforpython-6.2/PySide6/QtGui/QStandardItem.html#PySide6.QtGui.PySide6.QtGui.QStandardItem.icon) is built on PySide6. Then the Pixmap obtained is also PySide6 native. The new change in this PR forces it to use _our_ patched version of `QPixmap` and the `isinstace` is not failing anymore.

Let me know if this makes sense now.

python/shotgun_fields/shotgun_field_delegate.py
```diff
@@ -505,7 +505,9 @@ def _set_widget_value(self, widget, model_index):
             primary_item = src_index.model().item(src_index.row(), 0)
             icon = primary_item.icon()
             if icon:
-                widget.set_value(icon.pixmap(QtCore.QSize(256, 256)))
+                # Construct a pixmap object using the available QPixmap class
+                # When running in PySide6, the QPixmap class will be the already patched
+                widget.set_value(QtGui.QPixmap(icon.pixmap(QtCore.QSize(256, 256))))
```
staceyoue, line 510: I would patch the `QIcon.pixmap` method in tk-core to always create the pixamp from the patched QPixmap class. This way we do not need to remember to convert the native QPixmap to our patched class every time we use QIcon.pixmap

staceyoue (changes requested):

staceyoue:
> @staceyoue @julien-lang turns out the standard model's [icon](https://doc.qt.io/qtforpython-6.2/PySide6/QtGui/QStandardItem.html#PySide6.QtGui.PySide6.QtGui.QStandardItem.icon) is built on PySide6. Then the Pixmap obtained is also PySide6 native. The new change in this PR forces it to use _our_ patched version of `QPixmap` and the `isinstace` is not failing anymore.
> 

Great that we found where the native QPixmap object is getting created. See my code comment to move this change to tk-core.

Also, I noticed you closed this [PR](https://github.com/shotgunsoftware/tk-core/pull/975) to check qt instances. I think this PR would still be very helpful to complete because even though we have solved it for this particular case, there is likely to be other cases where the native Qt object is created instead of our patched class, and we will hit this issue again.

carlos-villavicencio-adsk:
We'll discuss and prioritize the implementation of the custom isinstance helper method. It's a bit out of the scope for this ticket but I'm sure we can work on it later.

python/shotgun_fields/status_list_widget.py
```diff
@@ -78,7 +78,7 @@ def setup_widget(self):
 
         self.activated.connect(lambda i: self.value_changed.emit())
 
-        self.setSizeAdjustPolicy(self.AdjustToContents)
+        self.setSizeAdjustPolicy(QtGui.QComboBox.AdjustToContents)
```
julien-lang, line 81: why do we need this change?

carlos-villavicencio-adsk: It seems these constants were part of the `QComboBox` instances on PySide (1). However, since PySide2 it's recommended to use the [class](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QComboBox.html#PySide2.QtWidgets.PySide2.QtWidgets.QComboBox.sizeAdjustPolicy) directly. In PySide6, this was completely deprecated.

There were other similar changes when supporting to PySide6 but I cannot find any other example at the moment.

julien-lang: Ok, I still don't understand why this would not work. I mean the current class inherits from `QtGui.QComboBox` so `self` would have access to the same constants

julien-lang (approved):
LGTM!

staceyoue (approved):

PR #163, opened 2024-05-22
SG-30560 Code cleanup for qt4 and pyside

eduardoChaucaGallegos:
Code cleanup for qt4 and pyside

### DIFF
--- a/python/elided_label/elided_label.py
+++ b/python/elided_label/elided_label.py
@@ -16,8 +16,6 @@ correctly within the widget frame.  Handles rich-text.
 import sgtk
 from sgtk.platform.qt import QtCore, QtGui
 
-utils = sgtk.platform.import_framework("tk-framework-shotgunutils", "utils")
-
 
 class ElidedLabel(QtGui.QLabel):
     """
@@ -208,7 +206,7 @@ class ElidedLabel(QtGui.QLabel):
 
             self._line_width = line_width
             return doc.toHtml()
-        except Exception:
+        except:
             pass
 
     @property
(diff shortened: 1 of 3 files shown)


julien-lang (commented):
LGTM!

carlos-villavicencio-adsk (approved):

staceyoue (approved):

julien-lang (approved):
LGTM!

python/activity_stream/activity_stream.py
```diff
@@ -28,7 +28,6 @@
 shotgun_globals = sgtk.platform.import_framework(
     "tk-framework-shotgunutils", "shotgun_globals"
 )
-utils = sgtk.platform.import_framework("tk-framework-shotgunutils", "utils")
```
julien-lang, line 31: Is this framework still used at all?

python/elided_label/elided_label.py
```diff
@@ -210,8 +206,8 @@ def _elide_text(self, text, elide_mode):
 
             self._line_width = line_width
             return doc.toHtml()
-        finally:
-            utils.safe_delete_later(doc)
+        except:
+            pass
```
julien-lang, line 210: What's the point? Can you remove the try block instead?

python/activity_stream/activity_stream.py
eduardoChaucaGallegos: In this repository, it was only used to call "utils.safe_delete_later()". I also reviewed if this function is used in other repositories but is used only here. The reason that it is deleted is that this function was deleted from "tk-framework-shotgunutils" also as part of Code cleanup for qt4

julien-lang: Yes, that makes sense. 
My question was whether or not _tk-framework-shotgunutils_ was used somewhere else in qtwidgets. Because if not, we would cleanup the "dependency" in `info.yml`.
But I checked and yes,  _tk-framework-shotgunutils_  is still massively used in this repo so no need no cleanup.

PR #166, opened 2024-07-02
SG-35935 first phase python2 dropping

eduardoChaucaGallegos:
- first phase python2 dropping

### DIFF
--- a/python/activity_stream/data_manager.py
+++ b/python/activity_stream/data_manager.py
@@ -10,15 +10,19 @@
 
 from sgtk.platform.qt import QtCore, QtGui
 
-import sgtk
 import copy
-import time
-import os
-import sys
-from tank_vendor import sgutils
 import datetime
-import sqlite3
 import hashlib
+import sgtk
+import sqlite3
+import sys
+import os
+import time
+
+try:
+    from tank_vendor import sgutils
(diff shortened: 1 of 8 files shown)


carlos-villavicencio-adsk (approved):

python/activity_stream/data_manager.py
```diff
@@ -10,15 +10,19 @@
...
-from tank_vendor import six
 import datetime
-import sqlite3
 import hashlib
+import sgtk
+import sqlite3
+import sys
+import os
+import time
+
```
julien-lang, line 21: Can we move the "sgtk" related imports below these ones?
```
<python builtin imports>

<project related imports>

<local imports>
```

This is just a personal preference

julien-lang (approved):

python/activity_stream/data_manager.py
carlos-villavicencio-adsk: A valid preference. It's is aligned with [PEP 8](https://peps.python.org/pep-0008/#imports).

>Imports should be grouped in the following order:
>
> 1. Standard library imports.
> 1. Related third party imports.
> 1. Local application/library specific imports.
>
> You should put a blank line between each group of imports.

PR #167, opened 2024-07-09
SHOT-4246: Improve note input widget UI

staceyoue:
* Make close/submit buttons more obvious by replacing icon buttons (to the right side of the note editor) with "Cancel"/"Submit" text buttons placed under the note editor
* Replace attachment close/submit buttons to be consistent with the new note cancel/submit buttons
* Moved "hint label" as placeholder text in the note editor
* Added more informative tooltips

**Note:**
    * Added `SGSubmitPushButton` and `SGCancelPushButton` classes to `sg_qwidgets` to use across Toolkit Apps to start having a more consistent UI across Toolkit and with the Web

3ds Max current:
3dsmax-panel-note-current
3dsmax-panel-reply-current
3dsmax-panel-attach-current

3ds Max with this code change:

3ds-max-note

3ds-max-reply

3ds-max-reply-attach

For reference, this is what the note input widget looks like from FPT Web:
image


python/note_input_widget/ui/note_input_widget.py
```diff
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
 
-# Form implementation generated from reading ui file 'note_input_widget.ui'
+# Form implementation generated from reading ui file 'C:\Users\qa\sg_envs\dev\tk\tk-framework-qtwidgets\python\note_input_widget\resources\note_input_widget.ui'
```
julien-lang, line 3: That's not the first time I see that in a PR. Did something changed in our Qt res scripts that would force an absolute path? 

We are updating the script in shotgunsoftware/tk-core#977 so if there is a problem, we should address it there.

python/note_input_widget/ui/note_input_widget.py
```diff
@@ -30,69 +31,70 @@ def setupUi(self, NoteInputWidget):
...
-        self.button_layout_right.setObjectName("button_layout_right")
-        self.close = QtGui.QToolButton(self.note_editor_page)
+        self.verticalLayout.addWidget(self.text_entry)
+        self.horizontalLayout_5 = QtGui.QHBoxLayout()
+        self.horizontalLayout_5.setSpacing(7)
+        self.horizontalLayout_5.setContentsMargins(-1, 3, -1, 3)
+        self.horizontalLayout_5.setObjectName("horizontalLayout_5")
+        self.screenshot = QtGui.QToolButton(self.note_editor_page)
         icon = QtGui.QIcon()
-        icon.addPixmap(QtGui.QPixmap(":/tk_framework_qtwidgets.note_input_widget/close.png"), QtGui.QIcon.Normal, QtGui.QIcon.Off)
```
julien-lang, line 54: Do we still use close.png anywhere? If not, we could cleanup this resource from the repo.

python/note_input_widget/ui/note_input_widget.py
```diff
@@ -145,77 +146,61 @@ def setupUi(self, NoteInputWidget):
...
+        self.add_button.setIcon(icon2)
         self.add_button.setAutoRaise(True)
         self.add_button.setObjectName("add_button")
         self.horizontalLayout_4.addWidget(self.add_button)
         self.remove_button = QtGui.QToolButton(self.attachment_list)
-        icon5 = QtGui.QIcon()
-        icon5.addPixmap(QtGui.QPixmap(":/tk_framework_qtwidgets.note_input_widget/minus.png"), QtGui.QIcon.Normal, QtGui.QIcon.Off)
-        self.remove_button.setIcon(icon5)
+        icon3 = QtGui.QIcon()
+        icon3.addPixmap(QtGui.QPixmap(":/tk_framework_qtwidgets.note_input_widget/minus.png"), QtGui.QIcon.Normal, QtGui.QIcon.Off)
```
julien-lang, line 161: What are the plus/minus buttons for? When reviewing the before/after screenshots, I was confused with these.

python/sg_qwidgets/__init__.py
```diff
@@ -27,6 +27,8 @@
     SGQMenu,
     SGQProgressBar,
     SGQPushButton,
+    SGSubmitPushButton,
+    SGCancelPushButton,
```
julien-lang, line 31: Minor question: why do we define so many things in our `__init__.py`? Would that not be easier (and less code to maintain) to keep it (almost) empty and use the regular Python import mechanism to import from sg_qwidgets?


This comment/question is not specific to this change but more like a general question of mine of the SGTK code. Like is there something I'm missing?

julien-lang (commented):
Looks good @staceyoue.

Just a few minor questions but no blocker. Not approving it because still a draft.

python/sg_qwidgets/sg_qwidgets.py
```diff
@@ -297,11 +297,100 @@ class SGQProgressBar(QtGui.QProgressBar):
...
+        qss = self.styleSheet()
+        qss += """
+            SGSubmitPushButton {
+                color: rgb(255, 255, 255);
+                background-color: rgb(24, 167, 227);
+                border: none;
+                border-radius: 4px;
+                padding: 4px 12px 4px 12px;
+            }
+        """
```
carlos-villavicencio-adsk, line 352: Personal question: why sometimes the styling of a button is done by qss code but the `__get_hover_effect` defined previously used class-based styling?

python/note_input_widget/ui/note_input_widget.py
staceyoue: I'm running on Windows, so I have my own powershell script to build resources...

python/note_input_widget/ui/note_input_widget.py
staceyoue: The plus/minus buttons only show when adding attachments to the note. You can use the "plus" icon to add an attachment, then you can select the attachment after it has been added and click the "minus" icon to remove it.

python/note_input_widget/ui/note_input_widget.py
staceyoue: We do not. I will clean this up, along with the "check mark" icon that is no longer used either.

python/sg_qwidgets/sg_qwidgets.py
staceyoue: From my experience, it is not consistent across Toolkit, whether style is applied using qss stylesheets or programmatically.

In this case, I prefer using qss stylesheets because it is much easier/quicker to apply style to a button for simple attributes (e.g. color, border, etc.), otherwise we have to override and re-implement the button `paintEvent` method, which is more complicated and more work to do. Though I prefer qss stylesheets, there are some more sophisticated styles that cannot be achieved using qss - this particular button hover effect I could not find a way to achieve it with qss.

python/note_input_widget/ui/note_input_widget.py
julien-lang: Ok so why do we see the minus button with the plus one when there is no attachment (one of your screenshot above shows that). 
I know you did not introduce that here but would that be something to clean up here? Or am I missing something else?

python/sg_qwidgets/__init__.py
staceyoue: Good question. I'm not entirely sure why this was the approach taken when the framework was initially set up, when implementing this `sg_qwidgets` modules I just followed the other modules.

One reason that defining each class is beneficial is that you can restrict which classes the consumer has access to. If the framework is kind of trying to act like an "api", this makes it more clear which modules the consumer should use.

The framework modules can be imported by a Toolkit EAF using this method:
```
sg_qwidgets = sgtk.platform.current_bundle().import_module("sg_qwidgets")
```
, through this import mechanism the sg_qwidgets/__ init __.py cannot be empty, it must import any modules that you would want to access by:
```
sg_qwidgets.SGQPushButton
```

You could replace all the individual imports with a single line:
```
from .sg_qwidgets import *
```
but then as noted above, a consumer has access to all modules defined and imported in the `sg_qwidgets` file, e.g. QtCore, QtGui, etc.

This example is for the `sg_qicons` module, if we had imported like `from .sg_qicons import *`

image

python/note_input_widget/ui/note_input_widget.py
staceyoue: That is just how the UI is currently. We can probably disable the button, when there are no attachments to remove

staceyoue: @julien-lang see description with updated screenshots for "-" button disabled when no attachment is selected to remove

julien-lang (approved):

carlos-villavicencio-adsk (approved):
LGTM

PR #170, opened 2024-09-04
SG-35973 Don't use ensure_str to prevent crashing

carlos-villavicencio-adsk:
If a task doesn't have a status set on SG, or tags, or the PublishedFile has no tags, they will be recognized as `None` in the breakdown2 component.

> This only happens when the user manually changes the PublishedFile version to `0` for their own purposes.

It seems it was working before this change: https://github.com/shotgunsoftware/tk-multi-breakdown2/pull/65. Adding @staceyoue as reviewer here.

|--|--|
|image | image|
|--|--|

Using `six.ensure_str` breaks as it doesn't support a `None` type. This change attempts to fix this.

image

### DIFF
--- a/python/filtering/filter_item_widget.py
+++ b/python/filtering/filter_item_widget.py
@@ -195,7 +195,7 @@ class ChoicesFilterItemWidget(FilterItemWidget):
             layout.addWidget(icon_label)
 
         # Left-aligned filter value display text
-        name = six.ensure_str(self._display_name)
+        name = str(self._display_name)
         self.label = QtGui.QLabel(name)
         layout.addWidget(self.label)
 


staceyoue (approved):
LGTM! Thanks

julien-lang (commented):
I don't understand how this change can fix the issue here. How does `ensure_str` behaves differently from `str`?

carlos-villavicencio-adsk:
> I don't understand how this change can fix the issue here. How does `ensure_str` behaves differently from `str`?

It's because of [the implementation](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank_vendor/six.py#L896). If the input is not string or binary, it raises an exception. This includes `NoneType`.

julien-lang (approved):

eduardoChaucaGallegos (approved):

PR #171, opened 2024-09-16
SG-36193 Adds named filter support 

pscadding:
Adds support for creating more complex filters shown in the filter menu as a preset filter.
This mimics the web UI behaviour in part, although there is no ability to modify the filters from the UI, only toggle which one is used.

There is a separate PR for the panel app listed below and the changes here to the framework are implemented to allow support for the panel app to work. I don't believe any of the included changes would work against implementing this in other apps like loader2 or breakdown app 2, but there would be further work needed to support those apps such as:

- saving the state of the active preset filter
- handling docked filter menu mode.

I did do some work into getting it to work in the loader, and had implemented the state storing but I've removed it due to the loader implementation becoming more work on the loader side.

Panel app support is added here:
https://github.com/shotgunsoftware/tk-multi-shotgunpanel/pull/97

Screenshot 2024-10-01 at 13 38 20

I did look at making the preset buttons larger to match the other filter buttons but this became quite involved and in my opinion added unnecessary complexity for this particular use. From what I could tell I could either adjust the styling on all menu items, or I would need to create an manage a widget that gets applied to the menu so that I could style it individually.


python/filtering/filter_menu.py
```diff
@@ -201,20 +205,21 @@ def __init__(
...
             self.__clear_widget.setChecked(False)
             self.__clear_widget.setCheckable(False)
-            self.__clear_widget.setText("Clear All Filters")
+            self.__clear_widget.setText("Clear Filters")
             sizePolicy = QtGui.QSizePolicy(
                 QtGui.QSizePolicy.Maximum, QtGui.QSizePolicy.Maximum
             )
             self.__clear_widget.setSizePolicy(sizePolicy)
+            # TODO should it be clearing preset filters here?
             self.__clear_widget.clicked.connect(self.clear_filters)
```
pscadding, line 221: I've not seen an example of this bit of code being hit. I suspect I should update the connection to use a lambda that calls the clear_filters with the clear_preset_filters

staceyoue: This clear widget is used when the filters are docked, and its behavior should be the same as the "Clear All Filters" menu action in the main Filter menu

image

python/filtering/filter_menu.py
```diff
@@ -264,6 +268,9 @@ def __init__(
         # based on menu selection.
         self._active_filter = FilterItem.create_group(FilterItem.FilterOp.AND)
 
+        # Set the active preset filter name to None, indicating no preset filter is active.
+        self._active_preset_filter_name = None
```
staceyoue, line 272: Is there a reason to treat preset filters differently than any other filters in the menu?

I would rather include the preset filter as part of the `self._active_filter`, this way all filters are contained in one object and filtering is handled by the framework (instead of having each App implement its own preset filter handling, which will simplify your tk-multi-shotgunpanel PR).

You should be able to modify `get_current_filters` to append the preset filter to the list of filters, which will create an AND with all other filters currently applied, and all filtering will be handled for you by the framework

staceyoue (changes requested):
@pscadding Looks good, but I do have a comment about refactoring how preset filters are handled. Let me know if it makes sense, or if you need any more clarification on it.

python/filtering/filter_menu.py
barbara-darkshot: hey @staceyoue !

I'm curious to see how you would handle these filters from a front-end perspective? It leads me to some questions :)

- how/where would you define them? still in a hook/setting by app I guess?
- then, you would modify the load method of the model(s) in each app to make sure all the fields required by the filters are fetched from FPTR when querying for the entities?
- what about if the applied preset is using a condition on a field and the same field is used by the user in the filter menu? does the `get_current_filters` method (or another one) should handle priorities? (for example, first apply the preset filter, then apply the other filters?)
- or should the preset filter be applied and then the list of other filters be refreshed accordingly? I don't remember if they're managing the "blank" value for example

staceyoue: All good points!

So after digging into this more, the TLDR: because the preset filters need to be defined as "shotgun filters", then they do need to be backend-filters.

It would be nice to have the framework handle all the filtering functionality, instead of each App that wants preset filters to have to introduce their own logic to handle the filtering, but currently the framework filtering can only handle filters that are built from the model data (and not shotgun filters).

python/filtering/filter_menu.py
```diff
@@ -275,7 +282,7 @@ def active_filter(self):
     @property
     def has_filtering(self):
         """Get whether or not the menu has any active filtering."""
-        return bool(self._active_filter and self._active_filter.filters)
+        return bool(self._active_filter and self._active_filter.filters or self._active_preset_filter_name)
```
staceyoue, line 285: Even though these are equivalent, I would prefer to use parentheses here to make it more clear what the intention is
```suggestion
        return bool((self._active_filter and self._active_filter.filters) or self._active_preset_filter_name)
```

python/filtering/filter_menu.py
```diff
@@ -351,6 +366,61 @@ def get_filters_container(self):
...
+            # In this case, we should emit that the preset filter has changed.
+            active_preset_filter = self.get_active_preset_filter()
+            new_filter_matching_active_name = preset_filters[self._active_preset_filter_name]
+            if active_preset_filter != new_filter_matching_active_name:
+                self.preset_filter_changed.emit()
+            self._preset_filters = preset_filters
+            return
+
+        # No changes in active preset filter, just update the preset filters.
+        self._preset_filters = preset_filters
```
staceyoue, line 405: I don't see this parameter being used (by tk-multi-shogunpanel), can we remove it to simplify this function logic?
If the caller wants to set the active preset filter, they should call `set_active_preset_filter`
```suggestion
    def set_preset_filters(self, preset_filters, new_active_preset_filter_name=None):
        """
        Set the preset filters that can be toggled on/off in the menu.
        The preset filters are stored as a dictionary where the key is the preset filter name,
        and the value is the filter data in the format the SG API accepts.
        :param preset_filters: dict of the preset filters to set.
        """

        self._preset_filters = preset_filters
```

staceyoue (changes requested):
After more thought, the preset filters should remain as backend-filters (ideally the framework could handle this use case, but it needs a lot more work to do so).  

Other than that, this is looking good, just a couple minor comments.

python/filtering/filter_menu.py
pscadding: yeah the issue I have here is that the named filters aren't implemented for the breakdown app so I had no way to test the changes here. I agree that it would need to be there if it was used. Usually in the past we've not implemented bits if they are not used.
Maybe I can add a test around it, even if nothing actually uses the functionality in practice (at least for the time being).

python/filtering/filter_menu.py
```diff
@@ -201,20 +205,21 @@ def __init__(
...
             self.__undock_widget.clicked.connect(self.undock_filters)
-            # Clear All Filtesr button for dock widget
+            # Clear All Filters button for dock widget
             self.__clear_widget = SGQToolButton(self.dock_widget)
             self.__clear_widget.setObjectName(
                 "filter_menu_dock_widget_clear_all_filters_button"
             )
             self.__clear_widget.setChecked(False)
             self.__clear_widget.setCheckable(False)
-            self.__clear_widget.setText("Clear All Filters")
```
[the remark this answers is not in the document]
pscadding: Yeah I don't know what I was thinking when I made this change. Strange, but I agree it should be Clear All Filters.
I'll change it back.

python/filtering/filter_menu.py
pscadding: Yeah good call, I had implemented it for the loader, and maybe you are right I could have just called `set_active_preset_filter` after setting the filters. But since I didn't go through with implementing this feature in the loader I will remove the param as you suggest any way. Thanks!

That said it can't be completely simplified, because the preset filters could change from call to call. So I need to handle the situations where the active preset filter no longer exists, or it exists but has changed.

python/filtering/filter_menu.py
pscadding: yep good call, I agree that is just easier to read.

staceyoue (approved):
LGTM!

julien-lang (approved):

PR #184, opened 2026-08-11
Ensure contexts saved from other configs are not deserialised

jonahjnewton:
`ContextWidget`'s `_settings_recent_contexts_key` setting stores contexts across pipeline configs. This can cause the initialisation of various pipeline configurations during `ContextWidget` setup when `sgtk.Context.deserialize(serialized_context)` is called in `_get_recent_contexts`, including execuction of any custom `TankInit` hooks across configurations.

To fix, we check the serialised contexts to ensure they're in the current pipeline configuration before deserialising.

Cheers
Jonah J. Newton (Flying Bark Productions)


julien-lang:
Hi @jonahjnewton, thanks for your contribution 😀

Could you provide more information about the issue you are fixing here, please? A full use case with reproduction path would be great🙏.

